PR #10, opened 2026-02-06
Implement argument constraints for tool evals

andreban:
Adds support for flexible argument matching in tool evaluations using constraint operato rs.
Key changes:
- Implemented matchesArgument in src/matcher.ts supporting:
  - `$pattern` (Regex match)
  - `$contains` (Substring match)
  - Numeric constraints (`$gt`, `$gte`, `$lt`, `$lte`)
  - `$type` (Type assertions)
  - `$any` (Presence check)
- Added comprehensive tests in src/test/matcher.test.ts.
- Documented constraints in README.md


evals-cli/package.json
beaufortfrancois: Should we update version or not?

PR #42, opened 2026-02-27
Nicer debug output in CLI & HTML report

### LINKED ISSUE #39: [Feature Request] More verbose evals reports including model thoughts
Currently, when a WebMCP evaluation finishes, the report shows tool calls with arguments, PASS/FAIL status and user message. 

In case of PASS this is enough. For FAIL case developer needs to understand why model did what it did, what was the reasoning?

- The report must interleave the model’s `thought` blocks with the corresponding `WebMCP` tool calls and responses.
- Ensure that the reasoning occurring immediately prior to a failed action is highlighted as the primary diagnostic data
- A full history of the model's logic for the entire session, collapsed by default with the "Failure Step" expanded.

swissspidy:
* Nicer progress bar on CLI, with correct Ctrl+C handling
* `--debug` flag for more verbose CLI output
* Sorted object keys for easier comparison between expected vs actual function arguments
* HTML report file is automatically opened in the browser upon finish

Example of the HTML report:

Screenshot 2026-02-27 at 17 47 11

Fixes #39

### DIFF
--- a/evals-cli/src/bin/webmcpevals.ts
+++ b/evals-cli/src/bin/webmcpevals.ts
@@ -8,10 +8,14 @@ import { resolve } from "path";
 import * as dotenv from "dotenv";
 import { Eval } from "../types/evals.js";
 import { WebmcpConfig } from "../types/config.js";
-import { SingleBar } from "cli-progress";
+import ora from "ora";
+import chalk from "chalk";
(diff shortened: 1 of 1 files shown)


Kulikowski (approved):
LGTM!

PR #113, opened 2026-03-30
Fixing linkability for our demos

Kulikowski:
Switched React and Angular demos to use **Hash Routing** to fix deep link breakages when hosted on GitHub Pages (which does not natively support SPA fallback to `index.html`).

#### React Demos
-   **`hotel-chain`** & **`react-flightsearch`**:
    -   Replaced `BrowserRouter` with `HashRouter`.
    -   Removed complex `basename` auto-detection logic, as hash routing is relative to the `index.html` location and does not require path guessing.

#### Angular Demo
-   **`sport-shop-angular`**:
    -   Enabled hash location strategy by updating `provideRouter` with `withHashLocation()`.

This standardizes all routing demos to work out-of-the-box on static hosting environments like GitHub Pages.

### DIFF
--- a/demos/hotel-chain/src/App.tsx
+++ b/demos/hotel-chain/src/App.tsx
@@ -3,43 +3,27 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import { BrowserRouter, Routes, Route } from 'react-router-dom';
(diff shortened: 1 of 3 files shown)


beaufortfrancois (approved):
LGTM

PR #136, opened 2026-04-09
Reverting and fixing deployment file

Kulikowski:
Reverting and fixing deployment file to preserve /demos/ path


beaufortfrancois (approved):
LGTM with nit

Kulikowski:
I'll revert previous PR, we need to fix few imporant things first :)

PR #140, opened 2026-04-10
fix(react-flightsearch): switch from HashRouter to BrowserRouter

andreban:
Replaces `HashRouter` with `BrowserRouter` in the react-flightsearch demo so URLs use clean query parameters (`/results?origin=...`) instead of hash-based routing (`/#/results?origin=...`).

### DIFF
--- a/demos/react-flightsearch/src/App.tsx
+++ b/demos/react-flightsearch/src/App.tsx
@@ -5,7 +5,7 @@
 
 import { useMemo } from "react";
 import {
-  HashRouter as Router,
+  BrowserRouter as Router,
   Routes,
   Route,
   useSearchParams,


Kulikowski:
We've added HashRouter in https://github.com/GoogleChromeLabs/webmcp-tools/commit/f8d12c87e4d3d40b90c5435a4182e60f467567df to support deep links in SPA without workarounds (like 404.html being clone of index.html) for static hosting on GitHub Pages.

To simulate this situation you can do
`npm run build`
`npx http-server dist` 

and then navigating with deep link should yield 404 error,

I think HashRouter is good enough tradeoff for hosting demo on GitHub Pages (not ideal as url is not "clean", but we don't introduce any workarounds),

Please let me know what do you think :)

andreban:
> We've added HashRouter in [f8d12c8](https://github.com/GoogleChromeLabs/webmcp-tools/commit/f8d12c87e4d3d40b90c5435a4182e60f467567df) to support deep links in SPA without workarounds (like 404.html being clone of index.html) for static hosting on GitHub Pages.
> 
> To simulate this situation you can do `npm run build` `npx http-server dist`
> 
> and then navigating with deep link should yield 404 error,
> 
> I think HashRouter is good enough tradeoff for hosting demo on GitHub Pages (not ideal as url is not "clean", but we don't introduce any workarounds),
> 
> Please let me know what do you think :)

Makes sense. I'll cancel this PR.

PR #150, opened 2026-04-22
fix: topping button when WebMCP is not enabled

marcianoskate:
Update the topping buttons to call the right signature of `addToppings` when WebMCP is not enabled.


beaufortfrancois (approved):
Thanks for catching!

PR #153, opened 2026-04-24
Remove unregisterTool calls

beaufortfrancois:
Now that AbortSignal has reached Chrome Stable channel, we can remove unregisterTool calls.

See https://chromiumdash.appspot.com/commit/71866d873bb9e87303f9009a188d429027fdd7a2


andreban (approved):
LGTM

PR #169, opened 2026-05-11
Fix live demo link in README

dineshb-qa:
Updated the live demo link to the correct URL for the coffee shop demo.


Kulikowski:
Thanks!

Kulikowski (approved):

PR #175, opened 2026-05-14
Search product is now globally available tool

Kulikowski:
Making search product globally available to support more complex CUJs

### DIFF
--- a/demos/sport-shop-angular/src/app/services/webmcp.service.ts
+++ b/demos/sport-shop-angular/src/app/services/webmcp.service.ts
@@ -98,5 +98,40 @@ export class WebmcpService {
         return { success: true, message: "Cart opened." };
       }
     });
+
+    // 4. Search Product Tool
+    modelContext.registerTool({
+      name: "search_product",
+      description: "Search for products based on the query.",
+      inputSchema: {
+        type: "object",
+        properties: {
+          query: {
+            type: "string",
+            description: "Search term for products (e.g. 'soccer boots', 'running gear')."
+          },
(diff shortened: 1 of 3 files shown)


demos/sport-shop-angular/src/app/services/webmcp.service.ts
```diff
@@ -98,5 +98,40 @@ export class WebmcpService {
...
+        }
+      },
+      execute: (params: any) => {
+        const searchParams: any = {
+          q: (params && params.query) ? params.query : ''
+        };
+        if (params && params.category && params.category !== 'ALL') searchParams.category = params.category;
+        if (params && params.size && params.size !== 'ALL') searchParams.size = params.size;
+
+        this.router.navigate(['/search'], { queryParams: searchParams });
```
beaufortfrancois, line 132: Side note: Do we need to await for the navigation to occur to follow our best practices? Something like `await this.router.navigate...` before returning result?

demos/sport-shop-angular/src/app/services/webmcp.service.ts
```diff
@@ -98,5 +98,40 @@ export class WebmcpService {
...
       }
     });
+
+    // 4. Search Product Tool
+    modelContext.registerTool({
+      name: "search_product",
+      description: "Search for products based on the query.",
+      inputSchema: {
+        type: "object",
+        properties: {
```
beaufortfrancois, line 108: Did you compare declarative inputSchema and this one? Which one worked the best?

PR #178, opened 2026-05-19
Use Puppeteer instead of modelContextTesting Web API

beaufortfrancois:
Now that https://chromiumdash.appspot.com/commit/625b9b44df117fba5a37be32dcd684ccc4a616e5 has landed into Chrome 150.0.7853.0, we can fully switch to Puppeteer instead of relying of the modelContextTesting experimental Web API which will be removed soon.

This CL does several things:
- Use `gemini-3-flash-preview` instead of deprecated `gemini-2.5-flash`
- Update system prompt to make sure model doesn't hallucinate tools
- Replace `listTools` calls with `page.webmcp.tools` from Puppeteer
- Replace `executeTool` calls with `tool.execute` from Puppeteer
- Replace `getCrossDocumentScriptToolResult` calls with manual `document.querySelector('script[type="application/ld+json"]')?.textContent` call
- Update puppeteer-core required version
- Update required Chrome version to 150+

### DIFF
--- a/evals-cli/README.md
+++ b/evals-cli/README.md
@@ -80,12 +80,12 @@ With Ollama:
 node dist/bin/runevals.js --model=qwen3:8b --backend=ollama --tools=examples/travel/schema.json --evals=examples/travel/evals.json
 ```
 
-| Argument     | Required | Default            | Description                                                              |
-| ------------ | -------- | ------------------ | ------------------------------------------------------------------------ |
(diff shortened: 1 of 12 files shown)


beaufortfrancois:
I've tried both french-bistro and pizza-maker demos. Let me know if its work for the other demos you own.

Kulikowski (approved):
LGTM

PR #182, opened 2026-05-26
fix(demos/analytics-dashboard): resolve query tool only after React rendering commits

andreban:
Addresses the race condition in the WebMCP analytics-dashboard query tool. The issue happened because `flushSync` forced the DOM to update, but didn't wait for the browser to re-render.

By using a deferred promise queue that resolves inside a React on-commit `useEffect` hook, we guarantee the tool only completes execution after the browser has successfully performed layout/paint cycles on the updated DOM, avoiding stale visual screenshots for AI agents.

This also removes usage of `flushSync`, which the [React documentation marks as uncommon and that it's usage may hurt performance](https://react.dev/reference/react-dom/flushSync).

### DIFF
--- a/demos/analytics-dashboard/src/App.jsx
+++ b/demos/analytics-dashboard/src/App.jsx
@@ -3,7 +3,6 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 import { useState, useEffect, useMemo, useCallback, useRef } from "react";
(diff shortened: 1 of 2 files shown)


Kulikowski (approved):
LGTM

andreban:
CC @MiguelsPizza -> interesting finding in our of our tests on `flushSync`.

PR #194, opened 2026-05-28
Improving parameter handling and waiting for tool loads

Kulikowski:
Improving parameter handling and adding temp waiting for tool loads

### DIFF
--- a/evals-cli/src/backends/vercel.ts
+++ b/evals-cli/src/backends/vercel.ts
@@ -183,7 +183,11 @@ export class VercelBackend implements Backend {
                   );
                 }
               : undefined,
-            prepareStep: (_opts: any): any => {
+            prepareStep: async (_opts: any): Promise<any> => {
+              // TODO Investigate no delay
+              // Wait 500ms for any pending page transition or rendering to settle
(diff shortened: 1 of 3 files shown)


evals-cli/src/backends/vercel.ts
```diff
@@ -183,7 +183,11 @@ export class VercelBackend implements Backend {
                   );
                 }
               : undefined,
-            prepareStep: (_opts: any): any => {
+            prepareStep: async (_opts: any): Promise<any> => {
+              // TODO Investigate no delay
+              // Wait 500ms for any pending page transition or rendering to settle
+              await new Promise((r) => setTimeout(r, 500));
```
beaufortfrancois, line 189: @Kulikowski  I'd love to understand why this is needed as this was not needed before in https://github.com/GoogleChromeLabs/webmcp-tools/pull/178/changes#diff-59d0217c7f6be32acdf7d82bbddd4be3203159a6fb7ffe3294acbc9d3f7483af

Kulikowski: Added during debugging Puppeteer integration. Left TODO here as I think it shouldn't be required here

PR #195, opened 2026-06-01
Removing not needed timeout

Kulikowski:
Tested within several runs, makes no noticeable difference.

### DIFF
--- a/evals-cli/src/backends/vercel.ts
+++ b/evals-cli/src/backends/vercel.ts
@@ -183,7 +183,7 @@ export class VercelBackend implements Backend {
                   );
                 }
               : undefined,
-            prepareStep: async (_opts: any): Promise<any> => {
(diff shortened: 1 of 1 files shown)


evals-cli/src/backends/vercel.ts
```diff
@@ -184,10 +184,6 @@ export class VercelBackend implements Backend {
                 }
               : undefined,
             prepareStep: async (_opts: any): Promise<any> => {
```
beaufortfrancois:
```suggestion
            prepareStep: (_opts: any): any => {
```

beaufortfrancois (approved):
LGTM with suggestion

evals-cli/src/backends/vercel.ts
Kulikowski: Good point! Updating

PR #196, opened 2026-06-01
Evals cli reporting

Kulikowski:
Improving the way tool reports results. Next step is to work on additional trajectory visibility for easier debugging.

Changes done to sidecar:
openWhenHidden: true: 
It instructs the library to keep the HTTP connection open and active in the background, regardless of whether you are looking at the tab or have switched away to another window. Prevent restarting evals when user switches tab

checkAbort.abort()
Cancels the active request: It immediately terminates the active HTTP connection.
Disables the retry loop: @microsoft/fetch-event-source is designed to automatically reconnect/retry indefinitely whenever a connection closes. When we call .abort(), the library recognizes that the cancellation was intentional, halts its internal retry timers, and exits the loop completely.

Updated evals to have names for easier debugging and grouping them across runs (if we specify N runs, then we will have N entries under eval name).

Preview:
image

NOTE: Evals are grouped based on 'name'. Accidental name collisions will result in Evals results being grouped as if those are same evals (we might want to introduce hashing of the eval case if we find it problematic)

### DIFF
--- a/evals-cli/README.md
+++ b/evals-cli/README.md
@@ -125,7 +125,7 @@ Each test case (eval) in the JSON definitions files supports the following prope
 | -------------- | -------------------- | -------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
 | `name`         | `string`             | No       | Descriptive name of the evaluation case. If provided, multiple execution iterations of this case (or stability runs) will be grouped together in the HTML report under a single accordion case. |
 | `messages`     | `Message[]`          | Yes      | An array of user prompts or conversation turns (multi-turn agent scenarios).                                                                                                                    |
-| `expectedCall` | `ExpectedCallNode[]` | Yes/No   | The list of expected tool calls. Can be `null` if no tool calls are expected.                                                                                                                   |
+| `expectedCall` | `ExpectedCallNode[]` | No       | The list of expected tool calls. Can be `null` if no tool calls are expected.                                                                                                                   |
 
 ### Example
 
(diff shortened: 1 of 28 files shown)


Kulikowski:
@beaufortfrancois @swissspidy happy to hear your thoughts!

From my perspective its way more readable than having Test 1-100 and not grouping same eval across runs

evals-cli/src/types/evals.ts
```diff
@@ -33,6 +33,7 @@ export type ExpectedCallNode =
   | { ordered: ExpectedCallNode[] };
 
 export type Eval = {
+  name?: string;
```
beaufortfrancois, line 36: Can you add `name` as well for multi-turn evals tests in `src/test/vercel.test.ts`?

evals-cli/README.md
```diff
@@ -116,6 +117,42 @@ node dist/bin/serve.js --port=8080
...
+| -------------- | -------------------- | -------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `name`         | `string`             | No       | Descriptive name of the evaluation case. If provided, multiple execution iterations of this case (or stability runs) will be grouped together in the HTML report under a single accordion case. |
+| `messages`     | `Message[]`          | Yes      | An array of user prompts or conversation turns (multi-turn agent scenarios).                                                                                                                    |
+| `expectedCall` | `ExpectedCallNode[]` | Yes/No   | The list of expected tool calls. Can be `null` if no tool calls are expected.                                                                                                                   |
```
beaufortfrancois: Is it required or no? I would no.

beaufortfrancois (approved):
LGTM with nits

evals-cli/README.md
Kulikowski: Fixed in next commit

PR #201, opened 2026-06-02
Using Angular way of exposing WebMCP tools

Kulikowski:
# Overview
Upgraded the leather-bag demo storefront to leverage experimental Angular v22 Next features, migrating the application to the new WebMCP APIs and Signal Forms package + product catalog improvements.

## Key Changes

Upgraded all @angular/* packages to 22.0.0-rc.3 and TypeScript to ~6.0.2 in package.json.
Opted into the new reactive forms and tool loop runtimes using provideExperimentalWebMcpForms() and provideExperimentalWebMcpTools() in app.config.ts.

Rewrote product option controls to use @angular/forms/signals.

## WebMCP Tool Suite (Global & Route-Scoped)

Global Tools: Registered check_return_policy, search_store, and view_product.
Route-Scoped Search Tools: Added filter_results, get_search_results, and add_search_result_to_cart under the /search route.

## Catalog Categorization & Term-Based Search

Categorized all products inside products.json and expanded the catalog with new items for the Accessories, Fine Details, and Apparel categories.
Refactored the search filter to parse queries into separate search terms (e.g., query "artisan belt" now successfully matches "Artisan Dress Belt").

## Interesting code places:
demos/leather-bag/src/app/app.config.ts
demos/leather-bag/src/app/app.routes.ts
demos/leather-bag/src/app/pages/product/product.ts


beaufortfrancois:
@dgp1130 Do you see any improvements in this PR that is switching from "manual" WebMCP tools to using Angular as documented in https://next.angular.dev/ai/webmcp?

demos/leather-bag/src/app/pages/product/product.ts
```diff
@@ -23,10 +23,41 @@ export class ProductComponent implements OnInit {
...
+    },
+    {
+      // Opt-in to implicit WebMCP tool add_to_cart with custom schema and submission action
+      experimentalWebMcpTool: {
+        name: 'add_to_cart',
+        description: 'Add this premium leather bag to your shopping cart with chosen color and quantity',
+      },
+      submission: {
+        action: async (formValue) => {
+          if (this.product) {
```
dgp1130: What are we expecting to happen in the `this.product = undefined;` case? Currently it seems like the submission succeeds, but is a no-op? Should this be an error or something?

demos/leather-bag/src/app/pages/product/product.ts
```diff
@@ -35,8 +66,15 @@ export class ProductComponent implements OnInit {
           this.product = p;
```
dgp1130: A bit unrelated to WebMCP, but I suspect you could simplify this by using `withComponentInputBinding()` to get a signal of the `id` parameter and then use a `rxResource` to call `getProductBySlug`. `selectedImage` can probably be a `computed` over `product`. `linkedSignal` could work for the selected color and image, but I think those should probably be folded into the signal model to have a single source of truth.

Setting the default color is a bit more awkward since it's async. I'd personally just drop it for the demo, but if you really want to keep it, you could make an `effect` on the `rxResource` which assigns it. If you can make everything signals, you should be able to drop `this.cdr.detectChanges()` (and even if you keep it, you probably want `this.cdr.markForCheck()`). Might look something like:

```typescript
class MyComponent {
  // Set by `withComponentInputBinding()` from URL
  readonly id = input.required<string>();

  // Creates a signal which defaults to `undefined` and calls `getProductBySlug` and then
  // updates itself with the result.
  private readonly product = rxResource({
    params: () => ({productId: this.id()}),
    stream: ({params}) => this.productService.getProductBySlug(params.productId),
  });

  // Derive the default color based on the `product`.
  private readonly defaultColor = computed(() => {
    const product = this.product();
    if (!product) return '';

    const hasBrown = product.colors.some((c) => c.name === 'Brown');
    return hasBrown ? 'Brown' : (product.colors[0]?.name ?? '');
  });

  // Dervie the default image based on the `product`.
  private readonly defaultImage = computed(() => {
    const product = this.product();
    if (!product) return '';

    return product.images[0];
  });

  // Signal Form for cart options
  readonly model = signal({
    color: '',
    image: '',
    quantity: 1
  });

  constructor() {
    effect(() => {
      const product = this.product();
      if (!product) return; // No-op until we have a product.

      // Update model if user didn't overwrite those values already.
      if (this.model().color !== '') {
        this.model.update((model) => ({...model, color: this.defaultColor()}));
      }
      if (this.model().image !== '') {
        this.model.update((model) => ({...model, image: this.defaultImage()}));
      }
    });
  }
}
```

* https://angular.dev/guide/routing/common-router-tasks#getting-route-information
* https://angular.dev/ecosystem/rxjs-interop#using-rxresource-for-async-data
* https://angular.dev/guide/signals/linked-signal

TBH, it might be simpler to lift the `rxResource` into a parent component and just render a loading spinner until you have the result, then render this component with `product = input.required<Product>();` and make the asynchronicity a lot simpler.

If this is too involved for what you're interested in, feel free to ignore, just sharing some thoughts around generally improving the form readability.

demos/leather-bag/src/app/pages/product/product.ts
```diff
@@ -46,49 +84,35 @@ export class ProductComponent implements OnInit {
...
-      
-      if (event && event.respondWith) {
-        event.respondWith(Promise.resolve({ 
-          success: true, 
-          message: `Added ${qty} ${this.selectedColor} ${this.product.name} to cart` 
-        }));
-      }
-    }
+    submit(this.cartForm);
   }
```
dgp1130: `formRoot` already calls `preventDefault` and `submit` for you, so I think you can just delete this handler altogether.

demos/leather-bag/src/app/app.config.ts
```diff
@@ -1,9 +1,72 @@
...
 import { provideHttpClient } from '@angular/common/http';
+import { provideExperimentalWebMcpTools } from '@angular/core';
+import { provideExperimentalWebMcpForms } from '@angular/forms/signals';
 import { routes } from './app.routes';
 
 export const appConfig: ApplicationConfig = {
-  providers: [provideBrowserGlobalErrorListeners(), provideRouter(routes, withHashLocation()), provideHttpClient()],
+  providers: [
+    provideBrowserGlobalErrorListeners(),
+    provideRouter(routes, withHashLocation()),
```
dgp1130: You should include [withExperimentalAutoCleanupInjectors](https://next.angular.dev/api/router/withExperimentalAutoCleanupInjectors) or else WebMCP tools on routes will continue to be available after navigating away from a route.

See the note here: https://next.angular.dev/ai/webmcp#provide-tools-for-a-route

demos/leather-bag/src/app/pages/product/product.ts
```diff
@@ -46,49 +84,35 @@ export class ProductComponent implements OnInit {
...
+    }
   }
 
   decrementQuantity() {
-    if (this.quantity > 1) this.quantity--;
+    if (this.quantity > 1) {
+      this.quantity--;
+      this.model.update(m => ({ ...m, quantity: this.quantity }));
+    }
   }
```
dgp1130: With the `min` and `max` validators on `this.model().quantity`, I don' think you need these `if` statements. The model is allowed to be invalid and show error messages, while the strict validation happens as part of submission. Though you may want to show an error message in the UI.

```html
@if (cartForm.quantity().touched() && cartForm.quantity().invalid()) {
  <ul class="error-list">
    @for (error of cartForm.quantity().errors(); track error) {
      <li>{{ error.message }}</li>
    }
  </ul>
}
```

https://angular.dev/essentials/signal-forms#validation-and-state

demos/leather-bag/src/app/pages/product/product.ts
```diff
@@ -23,10 +23,41 @@ export class ProductComponent implements OnInit {
...
-  // Accorion State
+  // Accordion State
   detailsOpen = true;
   shippingOpen = false;
 
+  // Signal Form for cart options
+  readonly model = signal({
+    color: '',
+    quantity: 1
+  });
```
dgp1130: If it's not too much trouble, I think it would be very cool to show off the array support, given dynamic forms has been a challenge for the declarative form API. If you could buy multiple products of different colors or a similar use case, you should be able to express that in the signal and Angular should generate an array type from it.

```typescript
readonly model = signal({
  variations: [{color: 'Red'}, {color: 'Blue'}],
});
```

dgp1130 (commented):
Thanks for incorporating the new Angular WebMCP APIs, looks great overall! Just left a few general comments.

demos/leather-bag/src/app/pages/product/product.ts
Kulikowski: Added error message

demos/leather-bag/src/app/pages/product/product.ts
Kulikowski: Added withComponentInputBinding PTAL

demos/leather-bag/src/app/pages/product/product.ts
Kulikowski: Added ( removed :) )

demos/leather-bag/src/app/app.config.ts
Kulikowski: Added

demos/leather-bag/src/app/pages/product/product.ts
Kulikowski: Good point! Added

Kulikowski:
@dgp1130 Applied suggestions, please take a look :)

AI calling tool "add_to_cart" with {"variations":[{"color":"Mahogany","quantity":3},{"quantity":2,"color":"Black"}]}
Tool "add_to_cart" result: {"content":[{"type":"text","text":"Form submitted successfully."}]}

image

demos/leather-bag/src/app/pages/product/product.ts
```diff
@@ -1,94 +1,145 @@
...
   private productService = inject(ProductService);
   private cartService = inject(CartService);
-  private cdr = inject(ChangeDetectorRef);
 
-  product?: Product;
-  selectedColor = '';
-  selectedImage = '';
-  quantity = 1;
+  // Set by `withComponentInputBinding()` from URL parameter `:id`
+  readonly id = input<string>();
```
dgp1130: I think you can use `input.required<string>();` to remove `undefined` from the type since this component is always used in a route with `:id`.

demos/leather-bag/src/app/pages/product/product.ts
```diff
@@ -1,94 +1,145 @@
...
+    }
+    const hasBrown = p.colors.some(c => c.name === 'Brown');
+    const defaultColor = hasBrown ? 'Brown' : (p.colors[0]?.name || '');
+    return {
+      variations: [{
+        color: defaultColor,
+        quantity: 1,
+      }]
+    };
+  });
```
dgp1130, line 60: I think there's a minor UX bug here where:

1. User interacts with the form (changes a color, increments the quantity).
2. `productResource` resolves with the product.
3. The user's state is clobbered and replaced with the default state of `variations: [{color: 'Brown', quantity: 1}]`.

I think you want to do something along the lines of:

```suggestion
  private readonly rawModel = signal([{color: '', quantity: 1}]);

  readonly model = linkedSignal<{ variations: Array<{ color: string; quantity: number }> }>({
    source: rawModel,
    computation: (newModel, previous) => {
      const p = this.productResource.value();
      if (!p) {
        return previous.source;
      }
      
      // Don't clobber user's state if they changed something.
      if (newModel.length !== 1 && newModel[0].color !== '' && newModel[0].quantity !== 1) {
        return previous.value;
      }
      
      const hasBrown = p.colors.some(c => c.name === 'Brown');
      const defaultColor = hasBrown ? 'Brown' : (p.colors[0]?.name || '');
      return {
        variations: [{
          color: defaultColor,
          quantity: 1,
        }]
      };
    },
  });
```

See: https://angular.dev/guide/signals/linked-signal#accounting-for-previous-state

I suspect part of the issue here is the UX of setting the "default" value asynchronously, which is a bit awkward to express in a consistent way. A simpler solution might just be to disable the form until the `resource` resolves with `disabled(f, {when: () => !productResource.hasValue()});`. Then you can probably leave your `linkedSignal` as-is based on the assumption that the computation will never run after the user sets their own data.

https://angular.dev/guide/forms/signals/form-logic#conditional-disabling

dgp1130 (commented):
Looks great! Glad to see the array functionality works as expected. I see one UX bug, but otherwise it looks pretty good to me.

demos/leather-bag/src/app/pages/product/product.ts
Kulikowski: Added conditional disabling

Kulikowski:
Included requested changes

demos/leather-bag/src/app/layout/header/header.spec.ts
```diff
@@ -1,6 +1,17 @@
...
+import { provideRouter } from '@angular/router';
+
+try {
+  TestBed.initTestEnvironment(
+    BrowserTestingModule,
+    platformBrowserTesting()
+  );
+} catch {
+  // Already initialized
+}
```
dgp1130: You shouldn't need this, Angular's unit test builder automatically injects and runs this for you: https://github.com/angular/angular-cli/blob/be251999be4ac6ecc685347271db10ad3c04ee62/packages/angular/build/src/builders/unit-test/runners/vitest/build-options.ts#L119-L123

Did you run into a specific issue which was fixed by this?

demos/leather-bag/package.json
dgp1130, line 1: Just FYI, v22 was released as stable yesterday, so you can switch all these packages to just `^22.0.0` and that should work.

(Also if you didn't run `ng update`, it might be worth doing that in case there are any migrations you should be applying.)

demos/leather-bag/src/app/layout/header/header.spec.ts
Kulikowski: You are correct! Removed it from all the tests, tests are passing

demos/leather-bag/src/app/pages/product/product.ts
```diff
@@ -1,94 +1,146 @@
...
+      applyEach(f.variations, (v) => {
+        required(v.color, { message: 'Color selection is required.' });
+        min(v.quantity, 1, { message: 'Quantity must be at least 1.' });
+        max(v.quantity, 10, { message: 'Quantity cannot exceed 10.' });
+      });
+    },
+    {
+      experimentalWebMcpTool: {
+        name: 'add_to_cart',
+        description: 'Add this premium leather bag to your shopping cart with chosen variations (color and quantity)',
```
beaufortfrancois: This description seems overly specific. Is that intentional?

demos/leather-bag/src/app/pages/product/product.ts
```diff
@@ -1,94 +1,146 @@
...
+      },
+      submission: {
+        action: async (formValue) => {
```
beaufortfrancois, line 78: The result I get is `{"content":[{"type":"text","text":"Form submitted successfully."}]}`. Could we override it so that the invoking agent knows how much and which color was actually added to the cart?

beaufortfrancois: Same would apply to other form tools

demos/leather-bag/src/app/pages/product/product.ts
```diff
@@ -1,94 +1,146 @@
...
+          const product = this.productResource.value();
+          if (!product) {
+            throw new Error('Product not loaded');
           }
-          this.cdr.detectChanges();
-        });
+          const variations = formValue.variations().value();
+          for (const v of variations) {
+            this.cartService.addToCart(product, v.color, v.quantity);
```
beaufortfrancois, line 85: I don't see bottom toast popping up when using WebMCP tool. Is that intentional?

demos/leather-bag/src/app/pages/product/product.ts
```diff
@@ -1,94 +1,146 @@
...
+    const defaultColor = hasBrown ? 'Brown' : (p.colors[0]?.name || '');
+    return {
+      variations: [{
+        color: defaultColor,
+        quantity: 1,
+      }]
+    };
+  });
+
+  readonly cartForm = form(
```
beaufortfrancois, line 62: When I'm on http://localhost:4200/#/product/signature-satchel page and I'm asking "Add one mahogany to my cart", it doesn't call the `add_to_cart` tool. Instead here's what it's doing:

```
AI calling tool "search_store" with {"query":"mahogany"}
Tool "search_store" result: {"content":[{"type":"text","text":"Navigated search results for \"mahogany\"."}]}
AI calling tool "get_search_results" with {}
Tool "get_search_results" result: {"content":[{"type":"text","text":"No search results found for the current search criteria."}]}
AI calling tool "search_store" with {"query":"leather bag"}
Tool "search_store" result: {"content":[{"type":"text","text":"Navigated search results for \"leather bag\"."}]}
AI calling tool "filter_results" with {"colors":["Mahogany"]}
Tool "filter_results" result: {"content":[{"type":"text","text":"Filters successfully updated."}]}
AI calling tool "get_search_results" with {}
Tool "get_search_results" result: {"content":[{"type":"text","text":"No search results found for the current search criteria."}]}
AI calling tool "search_store" with {"query":"all products"}
Tool "search_store" result: {"content":[{"type":"text","text":"Navigated search results for \"all products\"."}]}
AI calling tool "get_search_results" with {}
Tool "get_search_results" result: {"content":[{"type":"text","text":"No search results found for the current search criteria."}]}
AI calling tool "search_store" with {"query":"bag"}
Tool "search_store" result: {"content":[{"type":"text","text":"Navigated search results for \"bag\"."}]}
AI calling tool "get_search_results" with {}
Tool "get_search_results" result: {"content":[{"type":"text","text":"Found 7 search results:\n\n[0] The Signature Satchel - $1250\nCategory: Bags\nFinish: Full Grain\nColors: Mahogany, Black, Cognac\nDescription: Hand-selected full-grain hides, vegetable tanned and meticulously stitched by master artisans in our historic workshop.\n---\n[1] The Artisan Weekender Tote - $895\nCategory: Bags\nFinish: Vegetable Tanned\nColors: Tan, Black, Walnut\nDescription: Constructed from one continuous piece of water-resistant calfskin leather, designed to age with a unique patina.\n---\n[2] Handmade Leather Bag - $450\nCategory: Bags\nFinish: Full Grain\nColors: Brown, Tan, Black, Navy\nDescription: Every stitch is a testament to our legacy. This bag is handcrafted by master artisans using techniques passed down through generations.\n---\n[3] Classic Satchel - $245\nCategory: Bags\nFinish: Full Grain\nColors: Brown, Black, Cognac\nDescription: A classic satchel perfect for daily use. Features brass hardware and a durable strap.\n---\n[4] Artisan Tote - $310\nCategory: Bags\nFinish: Full Grain\nColors: Black, Tan, Chestnut\nDescription: Minimalist large black leather artisan tote bag.\n---\n[5] Travel Duffel - $595\nCategory: Bags\nFinish: Pebbled\nColors: Cognac, Black, Walnut\nDescription: Large travel duffel bag in aged cognac leather.\n---\n[6] Urban Rucksack - $285\nCategory: Bags\nFinish: Full Grain\nColors: Burgundy, Black, Tan\nDescription: Modern leather backpack in deep burgundy.\n---"}]}
```

beaufortfrancois: I think it's because the `color` input schema should include details with `toolparamdescription` for instance. It is currently quite empty.
```
 { "type": "object", "properties": { "variations": { "type": "array", "items": { "type": "object", "properties": { "color": { "type": "string" }, "quantity": { "type": "number" } }, "required": [ "color" ] } } }, "required": [] }
```

beaufortfrancois (changes requested):

demos/leather-bag/src/app/pages/product/product.ts
Kulikowski: Should be more generic

Kulikowski: It made sense when we had just bags in this store

demos/leather-bag/src/app/pages/product/product.ts
Kulikowski: It usually happens when you just start from non-zero journey place. What works is:

User prompt: "Add currently visible item to the cart in mahogany variant"
AI calling tool "add_to_cart" with {"variations":[{"quantity":1,"color":"mahogany"}]}
Tool "add_to_cart" result: {"content":[{"type":"text","text":"Form submitted successfully."}]}
AI result: OK. I've added the mahogany variant of the item to your cart.

I had the same issues with other demos when trying to imitate multi-turn evals as separate ones

Kulikowski: That said I think it would be worth to add toolparamdescriptions

demos/leather-bag/src/app/pages/home/home.spec.ts
```diff
@@ -9,6 +11,10 @@ describe('Home', () => {
   beforeEach(async () => {
     await TestBed.configureTestingModule({
       imports: [Home],
+      providers: [
+        provideHttpClient(),
+        provideRouter([]),
+      ],
```
JeanMeche, line 16: The client is provided in root. 
```suggestion
      providers: [
        provideRouter([]),
      ],
```

demos/leather-bag/src/app/pages/cart/cart.ts
```diff
@@ -53,4 +53,10 @@ export class CartComponent implements OnInit {
     this.shipping = this.subtotal > 500 ? 0 : 50;
     this.total = this.subtotal + this.shipping;
   }
+
+  checkout() {
+    this.cartService.clearCart('Thank you for shopping with us!');
+    this.calculateTotals();
+    this.cdr.detectChanges();
```
JeanMeche: I know the pattern is pre-exisiting but this shouldn't be need if the props were signals. 
```suggestion
```

demos/leather-bag/src/app/services/product.spec.ts
```diff
@@ -1,13 +1,18 @@
...
+  let service: ProductService;
 
   beforeEach(() => {
-    TestBed.configureTestingModule({});
-    service = TestBed.inject(Product);
+    TestBed.configureTestingModule({
+      providers: [
+        provideHttpClient(),
+      ]
+    });
```
JeanMeche: nit
```suggestion
    TestBed.configureTestingModule({});
```

JeanMeche:
If it helps, I could provide a follow-up to update this project to modern angular standards.

Kulikowski:
> If it helps, I could provide a follow-up to update this project to modern angular standards.

That would be great! Let's work towards merging this and then let's continue work on this project :)

demos/leather-bag/src/app/pages/product/product.ts
Kulikowski: I think the execution callback might be outside Angular's NgZone, as we will be doing follow up with some rewrites here I would leave it for now and I'll sync with @JeanMeche about it.

demos/leather-bag/src/app/pages/product/product.ts
Kulikowski: I think it's hardcoded in the framework @dgp1130 ?

demos/leather-bag/src/app/pages/product/product.ts
Kulikowski: @dgp1130 - toolparamdescription provides important context on how to properly use params to the agent, let's talk about it

PR #202, opened 2026-06-02
Add Page Agent demo

beaufortfrancois:
Following https://chromium-review.googlesource.com/c/chromium/src/+/7880667, this PR adds a page agent demo that shows how a webpage can get tools from a cross-origin iframe and execute them in a chat session powered by Gemini.

Here are the interesting bits:

```js
async function getTools() {
  const iframeOrigin = new URL(iframe.src).origin;
  const tools = await document.modelContext.getTools({ fromOrigins: [iframeOrigin] });
  return tools;
}
```
```js
            const result = await document.modelContext.executeTool(tool, inputArgs);
```   
    
Screenshot 2026-06-02 at 13 20 00

### DIFF
--- a/demos/page-agent/index.html
+++ b/demos/page-agent/index.html
@@ -1,0 +1,59 @@
+<!--
+ Copyright 2026 Google LLC
+ SPDX-License-Identifier: Apache-2.0
+-->
+
+<!DOCTYPE html>
+<html lang="en">
+
+<head>
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>WebMCP Page Agent</title>
+    <link
+      rel="icon"
+      href="data:image/svg+xml,<svg xmlns=%22http://www.w3.org/2000/svg%22 viewBox=%220 0 100 100%22><text y=%22.9em%22 font-size=%2290%22>💬</text></svg>"
+    />
+    <link rel="stylesheet" href="styles.css">
+</head>
+
+<body>
+    <div id="app-container">
+        <div id="browser-container">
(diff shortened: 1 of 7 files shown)


demos/page-agent/index.html
```diff
@@ -0,0 +1,59 @@
...
+      href="data:image/svg+xml,<svg xmlns=%22http://www.w3.org/2000/svg%22 viewBox=%220 0 100 100%22><text y=%22.9em%22 font-size=%2290%22>💬</text></svg>"
+    />
+    <link rel="stylesheet" href="styles.css">
+</head>
+
+<body>
+    <div id="app-container">
+        <div id="browser-container">
+            <div id="address-bar">
+                <input type="text" id="url-input"
```
Kulikowski, line 24: nit: I would suggest to make this not editable

Kulikowski (approved):
LGTM with nit

Kulikowski:
For visibility for other contributors: as discussed with Francois - Locally it behaves like cross-origin, when deployed on GH Pages it becomes same-origin, so we can deploy it somewhere else as well (chrome.dev for example)

demos/page-agent/index.html
beaufortfrancois: It is actually handy to test different demos directly from there.

beaufortfrancois:
> For visibility for other contributors: as discussed with Francois - Locally it behaves like cross-origin, when deployed on GH Pages it becomes same-origin, so we can deploy it somewhere else as well (chrome.dev for example)

Deploying on http://chrome.dev/web-ai-demos/webmcp/page-agent would be great indeed.

PR #222, opened 2026-06-09
Add WebMCP Registry information to documentation

JeffreyChix:
Added WebMCP Registry section with details on domain submission and tool registration.

The key difference from other directory tools already listed (e.g. webmcp.cool) is that submission is fully self-serve — developers submit their domain directly via the web UI, verify ownership with a DNS TXT record, and are immediately listed with no manual approval or contact required. The open source SDK (@webmcp-registry/kit) also lets teams define tools in code with Zod schemas and push them from CI automatically, rather than registering by hand.


beaufortfrancois:
Thx!

PR #223, opened 2026-06-10
Use document.referrer instead of history.back

beaufortfrancois:
Using `location = document.referrer` instead of `history.back()` loads the previous page instead of showing the page from the cache which would contain previous form data.


andreban (approved):
LGTM

PR #249, opened 2026-07-01
evals-cli: allow overriding provider base URLs via env

yoavweiss:
*Agent-generated PR*

## What

Adds optional `OPENAI_BASE_URL`, `ANTHROPIC_BASE_URL`, and `GOOGLE_GENERATIVE_AI_BASE_URL` environment variables to the eval-cli. When set, they are passed through to the corresponding `@ai-sdk/*` provider as `baseURL`. When unset, behavior is unchanged — each SDK falls back to its default provider endpoint.

The env-var names match the conventions used by the Vercel AI SDK ecosystem, so users who already have them configured for other tooling get consistent behavior.

## Why

Teams behind a corporate LLM gateway, or ones self-hosting an OpenAI-compatible endpoint (Azure OpenAI, LiteLLM, vLLM, an internal proxy, etc.), currently have no way to point `evals-cli` at their gateway. The provider `baseURL` is hardcoded to each SDK’s default.

The Ollama path already uses `OLLAMA_HOST` this way; this change makes the three cloud providers symmetric.

Concrete use case: at Shopify we route all internal LLM traffic through an OpenAI-compatible gateway fronting Gemini, GPT, Claude, and other providers. With this change we can follow the [Chrome WebMCP evals guide](https://developer.chrome.com/docs/ai/webmcp/evals) and use `evals-cli` directly against the gateway.

The gateway advertises an OpenAI-compatible surface for every underlying provider, so a single env override + the OpenAI provider covers Gemini/GPT/Claude/GLM/etc.

## Scope

Just the three provider entries in `src/evaluator/models.ts`, plus a small note in `README.md` alongside the existing env-var block. No new dependencies, no new CLI flags, no behavioral change when the vars are unset.

## Testing

- `npm run build` — clean
- `npm test` — 39/39 pass
- `npm run lint` — no new warnings (one pre-existing warning in `mappers.ts` unrelated to this change)
- `oxfmt --check` — clean
- End-to-end: ran all our eval cases via a real corporate OpenAI-compatible gateway, confirmed the request lands on the override URL rather than `api.openai.com`.

---

Co-Authored-By: AI (Claude Opus 4.7 via Pi)

### DIFF
--- a/evals-cli/src/evaluator/models.ts
+++ b/evals-cli/src/evaluator/models.ts
@@ -14,13 +14,17 @@ export function getModel(config: Config | WebmcpConfig) {
   if (config.provider === "openai" || modelId.startsWith("openai:")) {
     const apiKey = process.env.OPENAI_API_KEY;
     if (!apiKey) console.warn("Warning: OPENAI_API_KEY is missing for OpenAI provider.");
-    return createOpenAI({ apiKey })(modelId.replace("openai:", ""));
+    return createOpenAI({ apiKey, baseURL: process.env.OPENAI_BASE_URL })(
+      modelId.replace("openai:", ""),
+    );
   }
 
   if (config.provider === "anthropic" || modelId.startsWith("anthropic:")) {
     const apiKey = process.env.ANTHROPIC_API_KEY;
     if (!apiKey) console.warn("Warning: ANTHROPIC_API_KEY is missing for Anthropic provider.");
-    return createAnthropic({ apiKey })(modelId.replace("anthropic:", ""));
+    return createAnthropic({ apiKey, baseURL: process.env.ANTHROPIC_BASE_URL })(
+      modelId.replace("anthropic:", ""),
+    );
   }
 
   if (config.provider === "ollama" || modelId.startsWith("ollama:")) {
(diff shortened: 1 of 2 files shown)


evals-cli/src/evaluator/models.ts
```diff
@@ -35,6 +39,9 @@ export function getModel(config: Config | WebmcpConfig) {
   const apiKey =
     process.env.GOOGLE_AI || process.env.GEMINI_API_KEY || process.env.GOOGLE_GENERATIVE_AI_API_KEY;
   if (!apiKey) console.warn("Warning: Missing Google/Gemini API key");
-  const google = createGoogleGenerativeAI({ apiKey });
+  const google = createGoogleGenerativeAI({
```
beaufortfrancois, line 42: I wonder if we should use this opportunity to switch to `createGoogle()`. See https://github.com/vercel/ai/blob/dc1eb8da90387ffdec30d473d712f3aa292731c4/packages/google/src/google-provider.ts#L165

beaufortfrancois (approved):
@Kulikowski Please review.

Kulikowski (dismissed):
LGTM

Kulikowski:
> Thanks for your pull request! It looks like this may be your first contribution to a Google open source project. Before we can look at your pull request, you'll need to sign a Contributor License Agreement (CLA).
> 
> View this [failed invocation](https://github.com/GoogleChromeLabs/webmcp-tools/pull/249/checks?check_run_id=84512763076) of the CLA check for more information.
> 
> For the most up to date status, view the checks section at the bottom of the pull request.

@yoavweiss Hello! PR looks good and can be merged after signing CLA :) Thanks !

PR #250, opened 2026-07-01
Show in-page agent in result page

beaufortfrancois:
This PR makes sure the in-page agent is visible and opened by default when user lands on the result french-bistro page as requested by Khushal.

image


Kulikowski (approved):
LGTM

PR #259, opened 2026-07-02
Update model and use view-transition auto

beaufortfrancois:
This PR improves french-bistro cross-document UX with faster model and view transitions


Kulikowski (approved):
LGTM

PR #263, opened 2026-07-06
evals-cli: treat omitted `arguments` as unconstrained in functionCallOutcome

yoavweiss:
*_AI generated - Claude Opus 4.7_*

### Problem

An eval author who writes

```json
{ "functionName": "get_cart" }
```

in `evals.json` is asserting only "the model must call `get_cart`", with no constraint on its arguments. But every OpenAI-compatible SDK surfaces a no-arg tool call as `args: {}`, not `args: undefined`, so today the matcher scores that call as:

```
Function Name   exp='get_cart'   act='get_cart'   PASS
Arguments       exp='null'       act='{}'         FAIL   ← spurious
```

The root cause is in `functionCallOutcome` (`evals-cli/src/utils.ts`), which unconditionally forwards `expected.arguments` — `undefined` when the key is omitted — into `matchesArgument`. `matchesArgument` treats `undefined` as a strict literal and returns false against any object. `null` (if written explicitly) behaves the same way, since `matchesArgument(null, {})` is false by design (existing test at `matcher.test.ts:16-21`).

### Fix

In `functionCallOutcome`, if `expected.arguments` is nullish (missing key or explicit `null`), skip the args check entirely. Matches the intuitive semantics of the eval JSON — "I only care about the function name" — and deliberately doesn't touch `matchesArgument`'s existing strict handling of `null` as a literal, which the matcher tests rely on.

`FunctionCall.arguments` becomes `object | null | undefined` — reflecting that the field is in fact optional in practice.

### Tests

New in `utils.test.ts` (8 cases, in a fresh `functionCallOutcome` describe block):

- Omitted `arguments` matches `{}` and any non-empty actual args.
- Explicit `arguments: null` behaves the same way.
- Explicit `arguments: {}` still enforces "no args" — `{}` matches `{}`, `{ x: 1 }` does not (regression guard for the strict case).
- Basic pass/fail for matching function names, mismatched names, and matching / mismatching args.

Before: 39/39 pass.  After: 47/47 pass.

### Real-world impact

Found while running a 10-case eval suite against a real Gemini/GPT-compatible endpoint: 3 of the 8 "failures" were entirely this bug (`get_cart`, `cancel_cart`, `manage_orders`). Pass rate went from **4/10 (40%) → 7/10 (70%)** with just this one-line semantic fix applied.

### DIFF
--- a/evals-cli/src/test/utils.test.ts
+++ b/evals-cli/src/test/utils.test.ts
@@ -5,7 +5,12 @@
 
 import { describe, it } from "node:test";
 import * as assert from "node:assert";
-import { evaluateExecutionTrajectory, sortObjectKeys, countExpectedCalls } from "../utils.js";
+import {
+  evaluateExecutionTrajectory,
+  functionCallOutcome,
+  sortObjectKeys,
+  countExpectedCalls,
+} from "../utils.js";
 import { ExpectedCallNode } from "../types/evals.js";
 import { ToolCall } from "../types/tools.js";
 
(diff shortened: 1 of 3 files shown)


yoavweiss:
^^ @beaufortfrancois @Kulikowski

Kulikowski (approved):
LGTM, just needs `npx oxfmt`

Kulikowski (approved):

PR #267, opened 2026-07-06
evals-cli: preserve Gemini `thought_signature` on OpenAI-compat multi-turn

yoavweiss:

### Problem

Vertex-backed Gemini (via any OpenAI-compat proxy that fronts Vertex — Google's own compat endpoint, Shopify's AI proxy, corporate LiteLLM instances, etc.) emits a `thought_signature` field on every tool_call:

```json
{
  "tool_calls": [{
    "id": "call-1",
    "function": { "name": "search_catalog", "arguments": "{...}" },
    "extra_content": { "google": { "thought_signature": "AY89a..." } }
  }]
}
```

Vertex rejects follow-up requests that don't echo it back:

```
Unable to submit request because function call `search_catalog` in the 2.
content block is missing a `thought_signature`.
```

`@ai-sdk/openai` doesn't know about the `extra_content` provider-extension field, so when it builds turn N+1's message history from the previous assistant response it drops the signature. The net effect: multi-turn tool-calling against any Gemini variant through an OpenAI-compat proxy 400s on the second turn.

I ran into this while pointing the newly-merged `executeLocalEvals` multi-step support (that PR you just landed) at Vertex Gemini through the Shopify AI proxy — every case in an 11-eval suite errored on turn 2 with that message. Confirmed via direct `curl` against `POST /v1/chat/completions` that echoing the signature back inside `tool_calls[N].extra_content.google.thought_signature` on the second turn makes the same request succeed — the fix is fully client-side.

### Fix

A tiny `fetch` wrapper (`evaluator/googleThoughtSignatures.ts`, ~130 lines with docs) that:

1. **Intercepts responses** and stashes any `thought_signature` values, keyed by their tool_call `id`, in a per-instance `Map`.
2. **Intercepts subsequent requests** and re-injects the stashed signatures into outgoing `tool_calls[]` inside assistant messages — but only for calls that don't already carry one, so a caller who explicitly provides `extra_content` wins.

Wired into `getModel` on the OpenAI-compat path. Auto-enabled when the model id looks Google-shaped (`openai:google:*` or `openai:*gemini*`); force on/off via `OPENAI_PRESERVE_GOOGLE_THOUGHT_SIGNATURES=1|0`. Pure OpenAI / Anthropic-via-compat / Ollama traffic pays zero cost — those responses never carry `extra_content`, so the stash stays empty and outgoing requests are unchanged.

### Scope

- **Non-streaming only.** `generateText` (which `executeLocalEvals` uses) is non-streaming; that covers the immediate need. Streaming (SSE) support would need SSE-aware buffering to peek at each chunk; deferred until there's a caller that needs it.
- **JSON responses only.** Non-JSON content types (`text/event-stream`, etc.) pass through without inspection.
- **Per-instance state.** A fresh `Map` per `makeSignaturePreservingFetch()` call, so signatures don't leak across independent eval runs.

### Verification

Against a real Vertex-fronted proxy, my 11-case suite:

- **Before:** all 11 cases errored (`AI_APICallError: Bad Request` on turn 2, silently swallowed as case-level ERRORs before your recent `catch` fix)
- **After:** 0 framework errors, multi-step trajectories complete cleanly

Remaining suite failures are unrelated eval-authoring quirks (Gemini over-explores when a mock returns `{}`, matcher exact-key-set strictness on args like `pagination`) — I'll open a separate issue about the matcher strictness.

### Tests

7 new tests in `googleThoughtSignatures.test.ts`:

- Passthrough when no signatures are in play
- Extract-from-response, inject-into-next-request round trip
- Respect a caller-provided signature (no overwrite)
- Per-tool_call-id stashing (parallel calls in one turn)
- Non-JSON responses (SSE) pass through cleanly
- Malformed request bodies don't throw
- Fresh `Map` per factory call (isolation across independent eval runs)

Existing 66/66 pass; 73/73 with the new ones.

### DIFF
--- a/evals-cli/src/evaluator/googleThoughtSignatures.ts
+++ b/evals-cli/src/evaluator/googleThoughtSignatures.ts
@@ -1,0 +1,129 @@
+/**
+ * Copyright 2026 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * Vertex-backed Gemini (via any OpenAI-compat proxy that fronts Vertex)
+ * emits a `thought_signature` field per tool_call and rejects follow-up
+ * requests that don't echo it back:
+ *
+ *     Unable to submit request because function call `X` in the 2. content
+ *     block is missing a `thought_signature`.
+ *
+ * The signature lives in a provider-extension field
+ * (`tool_calls[N].extra_content.google.thought_signature`) that
+ * `@ai-sdk/openai` doesn't preserve when it builds the next turn's
+ * message history from the previous assistant response. So without any
+ * intervention, multi-turn tool-calling against Gemini through an
+ * OpenAI-compat proxy 400s on the second turn.
+ *
+ * This is a client-side workaround: a `fetch` wrapper that
+ *   1. Intercepts responses and stashes `thought_signature` values,
+ *      keyed by their tool_call `id`.
+ *   2. Intercepts subsequent requests and re-injects the stashed
+ *      signatures into the outgoing `tool_calls[]` inside assistant
+ *      messages.
+ *
+ * The wrapper is scoped per model instance (a fresh Map per
+ * `makeSignaturePreservingFetch()` call), so it doesn't leak signatures
+ * across independent eval runs. Non-JSON responses (e.g. SSE from
+ * streaming APIs) pass through untouched — this only helps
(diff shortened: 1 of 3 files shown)


yoavweiss:
^^ @Kulikowski

Kulikowski (approved):
LGTM

Issue #198, opened 2026-06-01
[Feature Request] Add matcher for optional tool calls


Kulikowski:
Currently we are checking for strict tool calls, while allowing for unordered and ordered tool calls. For models that are very eager to request tool calls we need a way to say that certain tool call might happen, but it's optional - for example to get current search results - such information might be useful for an agent to return verbose information to the user, but might not be needed/critical to fulfill a user journey.

I am open for proposals on how to mark expectedTool calls as optional in evaluation files. One idea I have is to start it's "name" with "*" or be more explicit and have additional attribute "optional": true.

PR #270, opened 2026-07-06
evals-cli: subset-match arguments in `matchesRecursive`

### LINKED ISSUE #268: Consider relaxing `matchesRecursive`'s exact-key-set requirement in argument matching

# Consider relaxing `matchesRecursive`'s exact-key-set requirement in `$pattern`/args matching

## Problem

`evaluateExecutionTrajectory` scores a tool call as PASS only when its `arguments` match the expected pattern recursively. The recursive walk (`matcher.ts:143`) hard-requires that the expected and actual objects have **exactly the same key set** at every level:

```ts
function matchesRecursive(expected: any, actual: any): boolean {
  ...
  const keys1 = Object.keys(expected);
  const keys2 = Object.keys(actual);
  if (keys1.length !== keys2.length) {
    return false;
  }
  ...
}
```

In practice this means an eval that expresses "I care about `query`, I don't care about anything else" has no way to say so. Any extra keys in the model's actual call — `pagination`, per-tool metadata, provider-injected defaults — fail the match, even when the constrained key(s) matched perfectly.

## Where this bites

While working through a real 11-case suite I ended up debugging the same pattern over and over:

```
### Search catalog when the user names a product   [0/1]
Function Name   exp='search_catalog'                                  act='search_catalog'                             PASS
Arguments       exp='{ "catalog": { "query": { "$contains": "running shoes" } } }'
                act='{ "catalog": { "pagination": { "limit": 10 }, "query": "running shoes" } }'  FAIL
```

The `$contains` matches. The eval author's intent ("must call `search_catalog` with a query containing `"running shoes"`") is 100% satisfied. But the model added a `pagination` field that the eval didn't mention, and the matcher rejects.

The workaround eval authors reach for is `pagination: {"$any": true}`:

```json
"arguments": {
  "catalog": {
    "query": { "$contains": "running shoes" },
    "pagination": { "$any": true }
  }
}
```

But `$any` requires the key be **present** — so the eval now flips from "extra pagination fails" to "missing pagination fails". Different models add pagination inconsistently (GPT-5.1 adds it for some queries and not others; Gemini adds it always). No static author-time value of the eval matches both worlds.

Concretely, in a suite I'm tuning, ~4 of ~15 case failures across two different models trace to this exact tension. When the matcher's decision doesn't align with author intent, the workaround eats a chunk of authoring time and the resulting evals are brittle to model-behavior changes.

## Proposal

Introduce a **subset-match** semantic for `matchesRecursive`: an object in the *expected* pattern matches if all of its keys are present in *actual* with matching values, regardless of whether actual has additional keys. Missing-in-actual still fails; extras-in-actual are ignored.

### Rough shape

```ts
function matchesRecursive(expected: any, actual: any): boolean {
  if (expected === actual) return true;
  if (
    expected === null || actual === null ||
    typeof expected !== "object" || typeof actual !== "object"
  ) return false;
  if (Array.isArray(expected) !== Array.isArray(actual)) return false;

  // Arrays: keep strict length + positional match. Arrays are usually
  // meaningful sequences (line_items, selected_options, ...) so treating
  // "extra elements are OK" would surprise authors.
  if (Array.isArray(expected)) {
    if (expected.length !== actual.length) return false;
    for (let i = 0; i < expected.length; i++) {
      if (!matchesArgument(expected[i], actual[i])) return false;
    }
    return true;
  }

  // Objects: subset match. Every key in expected must exist in actual and
  // its value must match; extra keys in actual are allowed.
  for (const key of Object.keys(expected)) {
    if (!Object.prototype.hasOwnProperty.call(actual, key)) return false;
    if (!matchesArgument(expected[key], actual[key])) return false;
  }
  return true;
}
```

Two behavioral changes worth flagging:

1. **Objects become subset-match** — extra actual keys are OK. This is the change that unblocks the `pagination` case above.
2. **Arrays stay exact-length** — arrays are usually positional (`line_items`, `selected_options`, `messages`, ...), and silently accepting extras there would surprise authors much more often than help.

### Migration risk

I'd expect most existing evals to keep passing: any eval that used to specify every key exactly still does; any that used `$any` explicitly to add a "must be present" gate still gets that behavior. Evals that were relying on strict-key-set to *reject* extras would need to opt back in — but I'd argue that's rare and could be handled with a new `$strict: true` marker at the object level if anyone needs it.

I'd propose testing the change against the existing test suite first (67 currently, would need a few new cases covering the subset semantics explicitly) and see what falls out.

### Alternative: opt-in via a `$subset` marker

If the maintainer preference is to keep the current default and add subset semantics only where explicitly requested, a marker approach would work:

```json
"arguments": {
  "catalog": {
    "$subset": true,
    "query": { "$contains": "running shoes" }
  }
}
```

`matchesConstraint` would see `$subset: true` alongside regular keys, switch to subset semantics for the containing object, and skip the length check. More explicit but more verbose in every eval.

I'd lean toward making subset the default (Option 1) because it's what authors reflexively expect when they write "match this key" and don't mention others, but I'm happy to build whichever shape you prefer.

## Ask

Would you take a PR for this? If so:

1. **Subset as default (my preference)** — cleaner semantics, small blast radius.
2. **`$subset` opt-in marker** — additive, no risk to existing evals.
3. **Neither** — keep exact-key-set, document the workaround pattern more explicitly.

Happy to prototype either shape once there's directional buy-in.

yoavweiss:

Closes #268.

### Problem

`matchesRecursive` currently requires expected and actual objects to have exactly the same key set — `keys1.length !== keys2.length` (`matcher.ts:143`) short-circuits to false. In practice this means an eval that expresses "I care about `query`, I don't care about anything else" has no way to say so: any extra key the model adds (`pagination`, per-tool metadata, provider-injected defaults) fails the match, even when the constrained key(s) matched perfectly.

Actual example from my suite:

```
exp: { catalog: { query: { $contains: "hoodie" } } }
act: { catalog: { pagination: { limit: 5 }, query: "black hoodie" } }
     → FAIL
```

The workaround authors reach for is `pagination: { $any: true }`, but `$any` requires the key to be **present** — the eval then flips from "extra pagination fails" to "missing pagination fails". Different models add pagination inconsistently, so no static author-time value matches both worlds.

### Fix

Switch object matching to subset semantics per your comment on #268 ("relaxing matcher and making subset a default makes a lot of sense"). Every key in `expected` must exist in `actual` with a matching value; extras in `actual` are allowed. Arrays keep their existing strict length + positional match — those are typically meaningful sequences (`line_items`, `selected_options`, ...) where silently accepting extras would surprise authors more often than help.

The change in `matchesRecursive`:

- Delete the `keys1.length !== keys2.length` gate for objects.
- Add an explicit branch for arrays that keeps strict-length behavior (both must be arrays, same length, positional match).
- Iterate only expected's keys when walking the object.

### Consequences worth calling out

- **Existing evals that used `arguments: {}` to assert "must have no args"** now silently accept any args — the empty expected set imposes no constraints. The one test that documented the old strict reading as a regression guard (added in #263) is updated in this PR to reflect the new contract.
- **`$strict: true` marker** as a future opt-back-in to exact-key-set is easy to add if anyone hits a case that needs it. Not included here — you mentioned that direction in the issue and I agree it's worth waiting until someone has a concrete need.
- **Arrays are still strict-length.** `[{a:1}]` matches `[{a:1,b:2}]` because the element itself subset-matches, but `[1,2]` does not match `[1,2,3]`.

### Tests

7 new tests in `matcher.test.ts` under a new "subset object matching" describe block:

- Extra keys in actual at the top level (the motivating case)
- Extra keys at nested levels
- Missing expected key in actual still fails
- Wrong value on an expected key still fails
- Empty expected object acts as "any object accepted" (documents the new contract)
- Arrays stay strict on length
- Subset composes with nested `$pattern` / `$contains` constraints

Existing 73/73 tests still pass (one had its assertion updated per the `{}` semantic change above); 80/80 with the new ones.

### Real-world impact

I dropped the modified `matcher.ts` into a vendored copy and re-ran my 11-case eval suite (against both Gemini and GPT-5.1). Two entries that were XFAIL'd in the storefront-kit expected-failures config immediately flipped to XPASS purely from this change:

- Gemini's `List collections when the user is exploring` — model added `navigate: false`, eval expected `{}`
- GPT-5.1's `Navigate to a product page on request` — model added `pagination`, eval expected `{ query: ... }` only

Both models now green with 0 unexpected fails.

Companion PR for #198 (optional tool calls at trajectory level) is coming right after this one.

---
*Co-Authored-By AI (Claude Opus 4.7) via [Pi](https://github.com/badlogic/pi-mono)*

### DIFF
--- a/evals-cli/src/matcher.ts
+++ b/evals-cli/src/matcher.ts
@@ -137,12 +137,20 @@ function isConstraintObject(obj: any): boolean {
 
 /**
  * Recursively checks equality between two values.
- * If values are objects or arrays, it recurses into them.
- * Crucially, it calls `matchesArgument` for children, enabling nested constraints.
  *
- * @param expected The expected structure.
- * @param actual The actual structure.
- * @returns True if structures match recursively.
+ * Object matching is **subset-based**: every key present in `expected` must
+ * exist in `actual` with a matching value, but extra keys in `actual` are
+ * ignored. This means an eval that specifies only the args it cares about
+ * (e.g. `{ query: { $contains: "shoe" } }`) matches a real tool call that
+ * also happens to include model-added fields like `pagination`.
+ *
+ * Array matching stays **strict**: expected and actual must have the same
+ * length and match positionally. Arrays are typically meaningful sequences
+ * (`line_items`, `selected_options`, ...) where silently accepting extra
+ * elements would surprise authors more often than help.
+ *
+ * Recurses via `matchesArgument`, so nested constraints (`$pattern`,
+ * `$contains`, `$any`, ...) continue to work inside subset-matched objects.
  */
 function matchesRecursive(expected: any, actual: any): boolean {
   if (expected === actual) {
@@ -158,14 +166,25 @@ function matchesRecursive(expected: any, actual: any): boolean {
     return false;
   }
 
-  const keys1 = Object.keys(expected);
-  const keys2 = Object.keys(actual);
-
-  if (keys1.length !== keys2.length) {
+  // Arrays keep strict length + positional matching.
+  const expectedIsArray = Array.isArray(expected);
+  if (expectedIsArray !== Array.isArray(actual)) {
     return false;
   }
+  if (expectedIsArray) {
+    if (expected.length !== actual.length) {
+      return false;
+    }
+    for (let i = 0; i < expected.length; i++) {
+      if (!matchesArgument(expected[i], actual[i])) {
+        return false;
+      }
+    }
+    return true;
+  }
 
-  for (const key of keys1) {
+  // Objects use subset-match: extra keys in `actual` are allowed.
+  for (const key of Object.keys(expected)) {
     if (
       !Object.prototype.hasOwnProperty.call(actual, key) ||
       !matchesArgument(expected[key], actual[key])
--- a/evals-cli/src/test/matcher.test.ts
+++ b/evals-cli/src/test/matcher.test.ts
@@ -39,6 +39,83 @@ describe("matcher", () => {
     });
   });
 
+  describe("subset object matching", () => {
+    it("accepts extra keys in actual at the top level", () => {
+      // The motivating case: an eval that constrains `query` but doesn't
+      // mention `pagination` should still pass when the model adds
+      // pagination on its own.
+      assert.strictEqual(
+        matchesArgument(
+          { catalog: { query: { $contains: "shoe" } } },
+          { catalog: { query: "purple shoe", pagination: { limit: 5 } } },
+        ),
+        true,
+      );
+    });
+
+    it("accepts extra keys at nested levels", () => {
+      assert.strictEqual(
+        matchesArgument(
+          { catalog: { query: "shoe" } },
+          { catalog: { query: "shoe", pagination: { limit: 5 }, sort: "asc" } },
+        ),
+        true,
+      );
+    });
+
+    it("still fails when an expected key is missing from actual", () => {
+      // Subset relaxes extras-in-actual, not missing-keys-in-actual.
+      assert.strictEqual(matchesArgument({ query: "shoe" }, { pagination: { limit: 5 } }), false);
+    });
+
+    it("still fails when an expected key has the wrong value", () => {
+      // Extras allowed; wrong values not.
+      assert.strictEqual(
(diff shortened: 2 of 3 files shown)


Kulikowski (approved):
LGTM

PR #271, opened 2026-07-06
evals-cli: support optional tool calls in expected trajectories

### LINKED ISSUE #198: [Feature Request] Add matcher for optional tool calls

yoavweiss:

Closes #198.

### Problem

Trajectories where "certain tool calls might or might not happen depending on the model" have no way to express the optionality today. Every listed `FunctionCall` is required; skipping one is a FAIL row. That forces eval authors to write either the eager model's trajectory or the lazy model's trajectory, losing coverage of the other.

Concrete example from my suite: an "eager" model calls a summary/inspection tool between a search and an add-to-cart; a more direct model goes straight through. Both are legitimate trajectories. Today the eval has to pick one.

### Design

An `optional?: boolean` field on `FunctionCall`. When set:

- If the model **does** make the call, it matches normally and contributes a PASS row (or a FAIL row if args mismatch — args-matching semantics on optional calls are the same as on required calls).
- If the model **doesn't**, the node contributes nothing: no PASS row, no FAIL row, no consumed actual, no impact on the overall trajectory outcome.

```jsonc
{
  "expectedCall": [{
    "ordered": [
      { "functionName": "search_catalog", "arguments": { ... } },
      // Eager models call this between search and add-to-cart; direct
      // models don't. Either behavior is a legitimate trajectory.
      { "functionName": "get_current_results", "optional": true },
      { "functionName": "update_cart", "arguments": { ... } }
    ]
  }]
}
```

Chose `optional: true` over the `*`-name-prefix alternative you sketched because:

- More explicit — reads self-documenting in JSON.
- Composes naturally with other `FunctionCall` fields (`mockOutput`, `arguments`, ...).
- Doesn't collide with any legitimate function name pattern.
- Structurally symmetric with the `arguments?: object | null` field's existing "opt-in by presence" shape.

### Implementation

Three places in `utils.ts`, ~40 real lines of code:

- **`countExpectedCalls`** excludes optionals — they don't inflate the progress-bar total or the per-case denominator.
- **`matchExpectedNode`** returns `matches: true, consumed: 0` for an optional `FunctionCall` that would fail to match (either because actual is exhausted or because the next actual doesn't match). A matching optional consumes normally (greedy match).
- **`matchSimpleUnorderedGroup`** (bipartite match for `unordered:`) filters unmatched optionals out of the FAIL-row generation and computes `allMatched` against the required-only count.

`matchNestedUnorderedGroup` gets support for free via `matchExpectedNode`, with one small change to its return: `consumed` is now derived from the best backtracking result rather than the pre-computed `expectedTotalConsumed` (which is required-only after the `countExpectedCalls` change and would have been wrong when optionals fire).

### Tests

9 new tests in `utils.test.ts`, in a new "optional tool calls" describe block:

- Skip an unmatched optional in an ordered sequence (PASS with fewer rows).
- Match an optional in an ordered sequence when the model does emit it.
- Required nodes still FAIL when actual is truncated, even next to a trailing optional.
- Skip an unmatched optional inside a simple unordered group.
- Match an optional inside a simple unordered group.
- Skip an unmatched optional inside a nested unordered group (exercises `matchNestedUnorderedGroup`).
- Extras in actual that match no expected still FAIL (optional isn't a wildcard).
- Optionals excluded from `countExpectedCalls` at leaf level.
- Optionals excluded from `countExpectedCalls` inside ordered/unordered subtrees.

Existing 73/73 tests still pass; 82/82 with the new ones.

### Companion PR

This composes cleanly with the subset-match matcher change in #270 (subset args + optional trajectory nodes both relax the matcher in the same "author intent, not model quirks" direction). Both PRs are independent — either can land first.

### Non-goals

- **Not** adding `optional: true` support on `ordered:` / `unordered:` groups themselves. If we ever need "this whole sub-trajectory is optional" that'd be a separate design pass — it introduces backtracking questions that need thought. For now, groups can just contain optional leaves.
- **Not** exposing optional in browser-eval mode differently. `executeInBrowserEvals` uses the same matcher, so it gets the semantics for free.

---
*Co-Authored-By AI (Claude Opus 4.7) via [Pi](https://github.com/badlogic/pi-mono)*

### DIFF
--- a/evals-cli/src/test/utils.test.ts
+++ b/evals-cli/src/test/utils.test.ts
@@ -273,6 +273,163 @@ describe("evaluateExecutionTrajectory", () => {
   });
 });
 
+describe("optional tool calls", () => {
+  it("skips an unmatched optional in an ordered sequence", () => {
+    // Expected: [search, get_summary?, get_product].
+    // Actual: [search, get_product] — the summary step was skipped.
+    // Trajectory should PASS with just the two required rows.
+    const expected: ExpectedCallNode[] = [
+      {
+        ordered: [
+          { functionName: "search" },
+          { functionName: "get_summary", optional: true },
+          { functionName: "get_product" },
+        ],
+      },
+    ];
+    const actual: ToolCall[] = [
+      { functionName: "search", args: {} },
+      { functionName: "get_product", args: {} },
+    ];
+    const results = evaluateExecutionTrajectory(expected, actual);
+    assert.deepStrictEqual(
+      results.map((r) => ({ fn: (r.expected as any)?.functionName ?? null, outcome: r.outcome })),
+      [
+        { fn: "search", outcome: "pass" },
+        { fn: "get_product", outcome: "pass" },
+      ],
+    );
+  });
+
+  it("matches an optional in an ordered sequence when the model does emit it", () => {
+    // Same expected trajectory as above, but the model DID make the
+    // optional call. All three rows should PASS.
+    const expected: ExpectedCallNode[] = [
+      {
+        ordered: [
+          { functionName: "search" },
+          { functionName: "get_summary", optional: true },
+          { functionName: "get_product" },
+        ],
+      },
+    ];
+    const actual: ToolCall[] = [
+      { functionName: "search", args: {} },
+      { functionName: "get_summary", args: {} },
+      { functionName: "get_product", args: {} },
+    ];
+    const results = evaluateExecutionTrajectory(expected, actual);
+    assert.strictEqual(results.length, 3);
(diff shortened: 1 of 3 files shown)


Kulikowski (approved):
Nice! LGTM

PR #276, opened 2026-07-09
Add WebMCP polyfill and integrate it with some demos

beaufortfrancois:
- Add shared `webmcp-polyfill.js` implementing current WebMCP API shape in Chromium.
- Refactor ModelContext to extend EventTarget with ontoolchange event handler support.
- Implement `registerTool()` supporting AbortSignal cancellation, and `getTools()` supporting origin filtering options.
- Implement `executeTool()` supporting both imperative callback execution and declarative form-filling tools.
- Support pending execution on forms without `toolautosubmit`, capturing submit events and resolving with `e.respondWith()` values.
- Dynamically translate CSS pseudo-classes (`:tool-form-active`, `:tool-submit-active`) to class selectors to support visual styling.
- Prevent tool serialization failures over French Bistro SharedWorker by returning only name, description, and inputSchema.
- Integrate the polyfill in French Bistro, Page Agent, Pizza Maker, and Sport Shop demos.

## Safari
Screenshot 2026-07-09 at 12 37 29

## Firefox
Screenshot 2026-07-09 at 12 36 12

PR #278, opened 2026-07-09
Support cross-origin tool querying and execution in polyfill

beaufortfrancois:
Following up on https://github.com/GoogleChromeLabs/webmcp-tools/pull/276

- Add remote tool inspection via postMessage using WEBMCP_GET_TOOLS_REQUEST/RESPONSE.
- Add remote tool execution via postMessage using WEBMCP_EXECUTE_TOOL_REQUEST/RESPONSE.
- Update ModelContext getTools() to inspect other frames asynchronously.
- Update ModelContext executeTool() to execute remote tools via message passing if _isRemote is set.
- Filter tools returned by getTools() based on options.fromOrigins: same-origin tools are always included, plus tools matching listed origins.

### DIFF
--- a/demos/shared/webmcp-polyfill.js
+++ b/demos/shared/webmcp-polyfill.js
@@ -21,66 +21,90 @@
     }
 
     // 2. Declarative tools (forms) on this window
-    try {
-      const forms = win.document.querySelectorAll('form[toolname]');
-      for (const form of forms) {
-        const name = form.getAttribute('toolname');
-        const description = form.getAttribute('tooldescription') || '';
+    const forms = win.document.querySelectorAll('form[toolname]');
+    for (const form of forms) {
+      const name = form.getAttribute('toolname');
+      const description = form.getAttribute('tooldescription') || '';
 
-        // Build inputSchema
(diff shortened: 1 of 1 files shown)


demos/shared/webmcp-polyfill.js
```diff
@@ -21,68 +21,92 @@
...
+    return new Promise((resolve) => {
+      const requestId = Math.random().toString(36).substring(2);
+      const timer = setTimeout(() => {
+        window.removeEventListener('message', listener);
+        resolve([]);
+      }, 500);
+
+      const listener = (event) => {
+        const { data } = event;
+        if (data && data.type === 'WEBMCP_GET_TOOLS_RESPONSE' && data.requestId === requestId) {
```
Kulikowski, line 93: Maybe worth to check for source as well? event.source === win

Kulikowski (approved):
LGTM

demos/shared/webmcp-polyfill.js
beaufortfrancois: Good catch!

PR #280, opened 2026-07-10
Enhance evals reports workflow

maudnals:


maudnals:
Obsolete, will open new one.

PR #287, opened 2026-07-14
Expose `execute_batch` tool and introduce "Code Mode" for Page Agent

swissspidy:
This PR adds a secure, declarative **Code Mode** (batch tool execution engine) for WebMCP. It allows AI agents to bundle multiple tool invocations and resolve data dependencies between steps (using a `$ref:stepId.property` format) in a single turn without executing arbitrary stringified code (e.g. avoiding insecure `eval`/`Function` call execution).

### Key Additions & Changes

1. **Standalone Batch Tool (`demos/shared/webmcp-batch.js`)**
   - Implements the `execute_batch` tool that executes steps sequentially, resolving any `$ref:stepId.nestedProperty` values from previous step results.
   - Halts execution upon encountering the first error to keep the client state predictable.
   - Exported UMD-style module to support both browser injection and Node.js importing.

2. **Unit Tests (`demos/shared/webmcp-batch.test.js`)**
   - Added 4 test suites for property resolution, recursive object reference replacement, reference forwarding between steps, and failure halting using Node's built-in `node:test` runner.

3. **Page Agent Upgrade (`demos/page-agent/`)**
   - **UI Toggle**: Added a switch in the header to alternate between Normal Mode (direct tool declaration) and Code Mode (exposing only `execute_batch`).
   - **TS Declarations Generator**: Generates type declarations for all available iframe guest tools to inject as system instructions for the LLM.
   - **Execution Logging**: Formats and prints structured step-by-step executions (input arguments, resolved values, output/results) directly inside the chat window.
   - **Local Fallback**: Automatically executes batch steps locally if the iframe page does not natively expose the `execute_batch` tool.

4. **Documentation**
   - Updated the Page Agent `README.md` to document Code Mode, the JSON steps structure, reference formats, and benefits.

### Testing Locally
- **Unit Tests**: Run `node --test demos/shared/webmcp-batch.test.js`.
- **Demos**: Run `./build-demos.sh`, open `demos/page-agent/index.html` in your browser, switch to **Code Mode**, and send a multi-step prompt.

### Demo

1 tool call for the model, triggering 4 tool calls on the website:

Screenshot 2026-07-14 at 11 18 27

### DIFF
--- a/demos/page-agent/README.md
+++ b/demos/page-agent/README.md
@@ -34,13 +34,52 @@ When the Gemini model decides to call a tool, the agent uses `executeTool` to pe
 const result = await document.modelContext.executeTool(tool, inputArgs);
 ```
 
+### 3. Code Mode (Batch Tool Execution)
+
+Code Mode introduces an optimized way for AI agents to interact with WebMCP tools. Instead of calling multiple individual tools in separate conversational turns (which consumes more context tokens and increases latency), the agent generates a single declarative steps array that calls multiple tools sequentially, resolving data dependencies between them.
+
+The agent uses a single tool called `execute_batch` with the following parameters:
+
+```json
+{
+  "steps": [
+    {
+      "id": "step1",
+      "tool": "set_pizza_size",
+      "args": { "size": "Large" }
+    },
+    {
+      "id": "step2",
+      "tool": "set_pizza_style",
+      "args": { "style": "BBQ" }
+    },
+    {
+      "id": "step3",
+      "tool": "add_topping",
+      "args": {
+        "topping": "🍄",
+        "count": 5,
+        "size": "$ref:step1"
+      }
+    }
+  ]
+}
+```
+
+#### How it works:
+1. **TypeScript Definitions**: The agent harness generates a TypeScript declaration (`mcp` interface) containing type signatures and descriptions for all available tools on the page.
+2. **System Instruction**: The declaration is injected into the agent's system instructions, alongside instructions on how to use declarative steps.
+3. **Single Tool Exposure**: Only the `execute_batch` tool is exposed to the model.
+4. **Execution**: The agent generates a sequence of steps. Step parameters can reference outputs of previous steps using the string format `"$ref:stepId"` or `"$ref:stepId.nestedProperty"`.
+5. **Output**: The batch execution parses and executes the steps sequentially, resolving any dependencies, and returns a list of step execution outputs (inputs, outputs/results, errors).
+
 ## ✨ Features
 
 - **Dynamic Tool Discovery**: Automatically detects tools from any WebMCP-compatible URL entered in the address bar.
+- **Normal & Code Mode**: Toggle between standard tool-calling and Cloudflare-style "Code Mode" to see the difference in context efficiency and speed.
 - **Gemini Integration**: Uses the Gemini 3.5 Flash model to interpret user intent and map it to available tools.
 - **Cross-Origin Support**: Safely interacts with tools across different origins via the WebMCP protocol.
-- **Real-time Feedback**: Shows system messages when tools are being executed.
-
+- **Real-time Feedback**: Shows system messages and live console logs when tools or scripts are being executed.
 ## 🚀 Getting Started
 
 1. Open the [Live Demo](https://googlechromelabs.github.io/webmcp-tools/demos/page-agent).
--- a/demos/page-agent/script.js
+++ b/demos/page-agent/script.js
@@ -4,6 +4,7 @@
  */
 
 import { GoogleGenAI } from 'https://esm.sh/@google/genai';
+import { executeDeclarativeBatch } from '../shared/webmcp-batch.js';
 
 const setupContainer = document.getElementById('setup-container');
 const chatContainer = document.getElementById('chat-container');
@@ -91,75 +92,6 @@ function getTSType(schema) {
     case 'object': return 'any';
     default: return 'any';
   }
-}
-
-function resolveReferences(val, results) {
-  if (typeof val === 'string') {
-    if (val.startsWith('$ref:')) {
-      const refPath = val.slice(5);
-      return getNestedProperty(results, refPath);
-    }
-    return val;
-  }
-  
-  if (Array.isArray(val)) {
-    return val.map(item => resolveReferences(item, results));
-  }
-  
-  if (val !== null && typeof val === 'object') {
-    const resolved = {};
-    for (const [k, v] of Object.entries(val)) {
-      resolved[k] = resolveReferences(v, results);
-    }
-    return resolved;
-  }
-  
-  return val;
-}
-
-function getNestedProperty(obj, path) {
-  const parts = path.split('.');
-  let current = obj;
-  for (const part of parts) {
-    if (current === null || current === undefined) return undefined;
-    current = current[part];
-  }
-  return current;
-}
-
-async function executeDeclarativeBatch(steps, executeToolFn) {
-  const results = {};
-  const outputs = [];
-
-  for (const step of steps) {
-    const { id, tool, args } = step;
-    const resolvedArgs = resolveReferences(args, results);
-    
-    try {
-      const result = await executeToolFn(tool, resolvedArgs);
-      if (id) {
-        results[id] = result;
-      }
-      outputs.push({
-        id,
-        tool,
-        args: resolvedArgs,
-        success: true,
-        result: result
-      });
-    } catch (err) {
-      outputs.push({
-        id,
-        tool,
-        args: resolvedArgs,
-        success: false,
-        error: err.message || String(err)
-      });
-      break;
-    }
-  }
-  
-  return outputs;
 }
 
 async function executeBatchLocally(steps) {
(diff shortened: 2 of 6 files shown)


swissspidy:
Quick test with 4 prompts after each other:

1. "Add a large BBQ pizza with mushrooms and onions"
2. "Add a large BBQ pizza with pepperoni"
3. "Now a small pizza with pineapple"
4. "Let's do a medium pizza with bacon and corn"

Without code mode:

```
📊 Last turn metrics:
↳ Input (Billed): 9,813 tokens
↳ Output (Billed): 91 tokens
↳ Total: 9,904 tokens
↳ Roundtrips: 3
↳ Time elapsed: 3.34s
```

With code mode:

```
📊 Last turn metrics:
↳ Input (Billed): 5,820 tokens
↳ Output (Billed): 131 tokens
↳ Total: 5,951 tokens
↳ Roundtrips: 2
↳ Time elapsed: 2.52s
```

* ~40% token savings
* ~25% speedup
* Number of tool calls constantly at 1

Kulikowski:
It's very interesting pattern! I'll experiment with it for other demos that we have, I think it's very nice optimization for tools that can be bundled together (working on same state, not triggering tool change/navigation)

swissspidy:
Some additional thoughts after a 1:1 discussion:

- This currently works best for a subset of tasks, i.e., ones without side-effects such as navigations that affect the list of available tools. This could be communicated via tool annotations perhaps.
- A batch tool like this could be offered by the website itself _or_ the agent harness.
- The evals CLI could be adapted to run evals with both modes, to demonstrate batching effectiveness.
- Benefits: reduced token usage, increased speed, and ideally also increased accuracy/reliability.
- Worth sharing at https://github.com/webmachinelearning/webmcp, as it's useful input for the spec.
- Batching is also related to skills or custom (user-generated) workflows that cover multiple tool calls (e.g., this way a custom `make_my_favorite_pizza` tool could declaratively call `set_pizza_size`, `set_pizza_style`, and `add_topping` under the hood)

demos/page-agent/script.js
```diff
@@ -15,22 +15,234 @@ const saveKeyBtn = document.getElementById('save-key-btn');
...
+      if (schema.items) {
+        return `${getTSType(schema.items)}[]`;
+      }
+      return 'any[]';
+    case 'object': return 'any';
+    default: return 'any';
+  }
+}
+
+function resolveReferences(val, results) {
```
Kulikowski: Duplicated method in webmcp-batch.js - I think we can just use shared version? Same for getNestedProperty, executeDeclarativeBatch.

demos/shared/webmcp-batch.js
```diff
@@ -0,0 +1,136 @@
...
+  const results = {};
+  const outputs = [];
+
+  for (const step of steps) {
+    const { id, tool, args } = step;
+    const resolvedArgs = resolveReferences(args, results);
+    
+    try {
+      const result = await executeToolFn(tool, resolvedArgs);
+      if (id) {
```
Kulikowski, line 58: nit: 

```
let parsedResult = result;
if (typeof result === 'string') {
  try {
    parsedResult = JSON.parse(result);
  } catch {
    // Keep original string if not valid JSON
  }
}
if (id) {
  results[id] = parsedResult;
}
```

Kulikowski (approved):
LGTM with comments

demos/page-agent/script.js
swissspidy: Yeah good call, I turned it into a script module to easily reuse those.

For context, this is just a local fallback within the agent if the iframed page doesn't already offer `execute_batch` itself.

swissspidy:
Another demo:

Works great if someone goes crazy ordering a pizza 😄 

Screenshot 2026-07-15 at 10 18 14

demos/page-agent/script.js
```diff
@@ -4,6 +4,7 @@
  */
 
 import { GoogleGenAI } from 'https://esm.sh/@google/genai';
+import { executeDeclarativeBatch } from '../shared/webmcp-batch.js';
```
beaufortfrancois, line 7: How about we import it dynamically instead?

demos/page-agent/script.js
```diff
@@ -15,22 +16,165 @@ const saveKeyBtn = document.getElementById('save-key-btn');
 const logoutBtn = document.getElementById('logout-btn');
 const urlInput = document.getElementById('url-input');
 const iframe = document.getElementById('iframe');
+const codeModeCheckbox = document.getElementById('code-mode-checkbox');
 
 let ai, chat;
 
+codeModeCheckbox.addEventListener('change', () => {
+  chat = null;
+  chatWindow.innerHTML = '';
```
beaufortfrancois, line 25: I'd remove `chatWindow.innerHTML = '';` so that we can compare code mode and normal mode in the same session.

demos/page-agent/script.js
```diff
@@ -15,22 +16,165 @@ const saveKeyBtn = document.getElementById('save-key-btn');
...
+    if (tool.name === 'execute_batch') continue;
+    
+    if (tool.description) {
+      decl += `  /**\n   * ${tool.description.split('\n').join('\n   * ')}\n   */\n`;
+    }
+    
+    let paramsType = 'Record<string, unknown>';
+    if (tool.inputSchema) {
+      try {
+        const schema = typeof tool.inputSchema === 'string' ? JSON.parse(tool.inputSchema) : tool.inputSchema;
```
beaufortfrancois, line 48: `typeof tool.inputSchema === 'string'` will always be true from getTools if tool.inputSchema is true.

demos/page-agent/script.js
```diff
@@ -15,22 +16,165 @@ const saveKeyBtn = document.getElementById('save-key-btn');
...
+      tools: [{ functionDeclarations: [executeBatchDecl] }]
+    };
+  }
+
+  // Normal Mode
   const systemInstruction = [
     'You are an assistant embedded in a web page.',
     'CRITICAL RULE: Do not try to use other tools than the available ones.',
-  ];
+  ].join('\n');
```
beaufortfrancois, line 176: Why do you need this?

demos/page-agent/script.js
```diff
@@ -44,6 +188,7 @@ async function getConfig() {
   return { systemInstruction, tools: [{ functionDeclarations }] };
 }
 
+
```
beaufortfrancois, line 191: Why this extra line?

demos/page-agent/script.js
```diff
@@ -138,7 +297,35 @@ async function handleUserSubmit() {
             appendMessage('System', `⚙️ Executing tool: ${name}...`, 'tool-indicator');
             const tools = await getTools();
             const tool = tools.find((t) => t.name == name);
-            const result = await document.modelContext.executeTool(tool, inputArgs);
+            
+            let result;
+            if (name === 'execute_batch' && !tool) {
```
beaufortfrancois, line 302: Shall we also check `codeModeCheckbox.checked`?

demos/page-agent/script.js
```diff
@@ -138,7 +297,35 @@ async function handleUserSubmit() {
...
+              result = await executeBatchLocally(args.steps);
+            } else {
+              result = await document.modelContext.executeTool(tool, inputArgs);
+            }
+
+            if (name === 'execute_batch') {
+              const batchResult = typeof result === 'string' ? JSON.parse(result) : result;
```
beaufortfrancois, line 309: My reading seems to indicate that batchResult will always be a dict. Why this check then?

demos/shared/webmcp-batch.js
```diff
@@ -0,0 +1,141 @@
...
+      
+      const outputs = await executeDeclarativeBatch(steps, executeToolFn);
+      
+      const success = outputs.every(o => o.success);
+      return {
+        success,
+        outputs
+      };
+    }
+  }, { exposedTo: ['*'] }).catch(() => {});
```
beaufortfrancois, line 135: `exposedTo: ['*']` is not valid as only secure origins are allowed in the exposedTo list. It means registerTool fails silently and is not working at all. Why do we have this then?

demos/page-agent/README.md
```diff
@@ -34,13 +34,52 @@ When the Gemini model decides to call a tool, the agent uses `executeTool` to pe
 const result = await document.modelContext.executeTool(tool, inputArgs);
 ```
 
+### 3. Code Mode (Batch Tool Execution)
+
+Code Mode introduces an optimized way for AI agents to interact with WebMCP tools. Instead of calling multiple individual tools in separate conversational turns (which consumes more context tokens and increases latency), the agent generates a single declarative steps array that calls multiple tools sequentially, resolving data dependencies between them.
```
beaufortfrancois, line 39: Nit: I'd flag this as experimental. 

FYI I had played with a similar idea with https://github.com/GoogleChromeLabs/webmcp-tools/pull/69 before

swissspidy: Thanks for sharing that, I wasn't aware! I first started with this eval approach as well but pivoted for security reasons. I do find the idea interesting though at the harness-level and want to explore using something like https://github.com/sebastianwessel/quickjs for that.

PR #288, opened 2026-07-15
Preparing NPM release

Kulikowski:
This PR restructures evals-cli into a CLI package (webmcp-evals) for evaluating WebMCP schemas, tool calling, and multi-turn agentic trajectories against both static schema files and live browser web applications.

# CLI Architecture (commander)
Removed side-car UI and added unified command-line tool (webmcp-evals).
There are two distinct evaluation commands:
`local`: Evaluates static JSON schema definition files (-t/--tools & -e/--evals).
`browser`: Launches Puppeteer browser sessions to dynamically evaluate live WebMCP tools declared on web pages 
via document.modelContext.getTools().

# Multi-Format Reporter Suite (src/report/)
Aggregates single and multi-run evaluations into three output formats:
console: Terminal summary with pass/fail indicators.
json: Report logs for CI/CD ingestion (.evals/report-<timestamp>.json).
html: HTML dashboard visualizing detailed step trajectories, expected vs. actual call diffs, and aggregate pass rates.

Added additional test coverage for Vercel, Gemini, and Ollama backend implementations.
Unit tests for dynamic tool mappers (mapMessages, mapRawBrowserToolsToConfig).
Improved the way browser evals are executed


Kulikowski:
image

evals-cli/package.json
```diff
@@ -1,14 +1,22 @@
 {
-  "name": "webmcp-evals-cli",
+  "name": "webmcp-evals",
   "version": "0.1.0",
-  "description": "A command-line tool to run evals for WebMCP schema definitions",
+  "description": "Evaluation framework and CLI for WebMCP tools and agentic workflows",
   "license": "Apache-2.0",
+  "bin": {
+    "webmcp-evals": "dist/bin/webmcp-evals.js",
+    "webmcpevals": "dist/bin/webmcp-evals.js"
```
swissspidy: Remove this one?

Kulikowski: Moved alias

evals-cli/run_evals.sh
swissspidy, line 1: We could create a GitHub Actions workflow for this that we could run manually or only when working on related files.

Kulikowski: My thinking was to have this a local tool to run to check basic evals, GitHub actions workflows with non-determinism (inference) might be tricky, what do you think?

Maybe have it as simple local script now and then follow up with something?

swissspidy: It wouldn't be 100% reliable for seeing whether everything passes, true. But if everything is _failing_, we know that something most likely broke. So I would use it more as a sanity check for that. But yeah can be in a follow-up of course.

maudnals, line 1: Useful addition 👍

maudnals (approved):
LGTM!

swissspidy (approved):

PR #293, opened 2026-07-16
Evals: Bugfix for runs vs steps, and full LLM trajectory

maudnals:
Fixes #61 

* Fix bug where Runs and Steps were confused
* Enhance report UI: clarify Steps vs Runs, display pass status for each, mention run # in report summary
* Display available tools at each step
* Renaming in the code: TestCase, TestRun, TestStep

See screenshots.

@Kulikowski Questions for you:
* I've tested with sports shop, what's another good multi-step demo to test this with?
* The summary displays a pass rate % and for now I've kept each test case summary and run summary to display a pass ratio (n/n). Are we happy with that?
* (One observation: AFAICS the step count can vary even for a specific run and test case, because it depends on what tools the LLM will decide to call. This wasn't intuitive to me at first and also means that the eval count isn't constant, because its step-based. Would it make sense to clarify that in the UI or think of alternatives?)
* Issue #61 mentions "all interactions that happened with browser". This PR surfaces available tools, what do we need to add? Thinking process at each step?

---

image
image
image

### DIFF
--- a/evals-cli/src/backends/vercel.ts
+++ b/evals-cli/src/backends/vercel.ts
@@ -271,9 +271,10 @@ export class VercelBackend implements Backend {
             }
           }
 
-          const rawSteps = resultPayload.steps && resultPayload.steps.length > 0
-            ? resultPayload.steps
-            : stepsHistory;
+          const rawSteps =
+            resultPayload.steps && resultPayload.steps.length > 0
+              ? resultPayload.steps
+              : stepsHistory;
 
           const trajectory = rawSteps.map((step, idx) => ({
             text: step.text,
--- a/evals-cli/src/backends/vercel.ts
+++ b/evals-cli/src/backends/vercel.ts
@@ -268,13 +268,21 @@ export class VercelBackend implements Backend {
 
           if (trajectories.length === 0) {
             const response: any = { text: resultPayload.text };
-            const stepResult: TestResult = { test, response, outcome: "pass", trajectory };
+            const stepResult: TestResult = {
+              test,
+              response,
+              outcome: "pass",
+              trajectory,
+              runIndex: r + 1,
+              stepIndex: 1,
+            };
             testResults.push(stepResult);
             passCount++;
             if (onEvent) {
               onEvent({ type: "progress", testNumber: testCount, result: stepResult });
             }
           } else {
+            let stepIndex = 1;
             for (const traj of trajectories) {
               let response: any = traj.actual;
               if (!response && executedCalls.length === 0 && resultPayload.text) {
@@ -292,6 +300,8 @@ export class VercelBackend implements Backend {
                 response,
                 outcome: traj.outcome,
                 trajectory,
+                runIndex: r + 1,
+                stepIndex: stepIndex++,
               };
 
               testResults.push(stepResult);
@@ -313,6 +323,8 @@ export class VercelBackend implements Backend {
             test,
             response: null as any,
             outcome: "error",
+            runIndex: r + 1,
+            stepIndex: 1,
           };
           testResults.push(result);
           if (onEvent) {
--- a/evals-cli/src/backends/vercel.ts
+++ b/evals-cli/src/backends/vercel.ts
@@ -179,6 +179,9 @@ export class VercelBackend implements Backend {
           );
         }
 
+        const availableToolsPerStep: Array<Array<Tool>> = [];
+        const stepsHistory: any[] = [];
+
         try {
           const model = getModel(config);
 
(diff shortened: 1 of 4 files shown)


Kulikowski:
@maudnals 

1) Ideally we should have more evals for complex demos, let's talk about this - imho can be left out of the scope of this PR
2) Looks good!
3) Let's talk about it - I think it makes sense to clarify, especially when dealing with overly eager-tool-calling models
4) Tools available at each step is minimum needed for debugging. Reasoning is also included - what I would propose is to start addressing point 1) (more evals) and see where we see gaps for debugging

evals-cli/src/backends/vercel.ts
```diff
@@ -179,6 +179,9 @@ export class VercelBackend implements Backend {
           );
         }
 
+        const availableToolsPerStep: Array<Array<Tool>> = [];
+        const stepsHistory: any[] = [];
```
Kulikowski: I would introduce some type as we provide structure for it below and assume it:

Something ~

export interface TrajectoryStep {
  text?: string;
  reasoningText?: string;
  toolCalls?: any[];
  toolResults?: any[];
  availableTools?: Tool[];
}

evals-cli/src/backends/vercel.ts
```diff
@@ -212,19 +215,27 @@ export class VercelBackend implements Backend {
...
-              : undefined,
+            onStepFinish: (event) => {
+              if (config.debug) {
+                console.log(
+                  `[DEBUG] Step ${event.stepNumber || ""} finished (${event.finishReason}). Total Tokens: ${event.usage.totalTokens}`,
+                );
+              }
+              stepsHistory.push({
+                text: event.text,
+                reasoningText: event.reasoningText,
```
Kulikowski, line 226: Worth checking across different model providers if the structure is the same. It might be `reasoningText` or `reasoning`.

evals-cli/src/report/report.ts
```diff
@@ -58,15 +58,29 @@ export function renderReport(config: Config, testResults: TestResults): string {
...
+    testResults.results.map(
+      (r) =>
+        r.test.name ||
+        (r.test.messages[0] && r.test.messages[0].type === "message"
+          ? r.test.messages[0].content
+          : ""),
+    ),
+  );
+  const totalCases = caseNames.size;
+  const runs = Math.max(...testResults.results.map((r) => r.runIndex || 1), 1);
```
Kulikowski: nit: const runs = testResults.results.reduce((max, r) => Math.max(max, r.runIndex || 1), 1);

evals-cli/src/report/report.ts
```diff
@@ -343,13 +417,39 @@ function renderTrajectory(trajectory?: any[]): string {
...
               "</strong>";
 
-            if (step.text) {
+            const thoughts = step.reasoningText || step.text;
+            if (thoughts) {
               html +=
                 '<div><em class="text-xs font-semibold text-slate-500 uppercase tracking-wider block mb-1">Thoughts:</em>' +
                 '<pre class="whitespace-pre-wrap bg-slate-50 p-3 rounded-md text-sm text-slate-700 border border-slate-200 font-sans">' +
-                step.text +
+                thoughts +
```
Kulikowski: I think we might want to escapeHTML there if thoughts would contain "Ok, tool description says I should use it for numbers < 5" the '<' could break the layout.

Something to consider for other dynamic additions as well

Kulikowski (approved):
LGTM with comments

evals-cli/src/backends/vercel.ts
maudnals: in the [Vercel docs](https://ai-sdk.dev/docs/reference/ai-sdk-core/generate-text#reasoningtext#on-step-end.step-result.reasoning-text#on-step-end): `reasoningText` will work across model providers but will be `undefined` for non-thinking models, which are IMU unlikely for WebMCP usage?

Side note: We could think of updating our vercel package version soon as I'm seeing some deprecations e.g. `onStepFinish` -> `onStepEnd`

evals-cli/src/report/report.ts
maudnals: Definitely, added a basic HTML escaping function, WDYT? 
Right now we only escapeHtml at the last moment so we're unlikely to accidentally escape twice, but something to keep in mind for the future.

evals-cli/src/backends/vercel.ts
maudnals: Added new `TrajectoryStep` type

evals-cli/src/report/report.ts
maudnals: Done!

Kulikowski:
LGTM!

PR #294, opened 2026-07-16
Gemini and Ollama following Backend interface contract

Kulikowski:
Update `GeminiBackend` and `OllamaBackend` to return `LocalEvalResult` (`{ toolCalls, text }`), matching the `Backend` interface contract. 
Also fixes a `@google/genai` SDK console warning when tool calls are present by extracting text directly from candidate parts instead of accessing `response.text`.


swissspidy (approved):
Nice consolidation, seems to work well when testing.

PR #298, opened 2026-07-17
Refactor of evals cli

Kulikowski:
# Refactor

This PR refactors how browser evaluations are run. The main change is separating the browser orchestration logic (launching the browser, handling page navigation, running test suites, and scoring results) from the model-specific execution loops.

### Key Changes:

1. **Decoupled Browser Orchestration**:
   * Extracted browser setup, test loops, and result compiling from the Vercel backend into a generic orchestrator (`executeInBrowserEvals` in `src/evaluator/browserEvaluator.ts`).
   * The Vercel backend's `executeInBrowserEval` method is now simplified to only run a single agent execution step on a pre-allocated browser page.

2. **Dependency Injection**:
   * Created a `createBackend` factory in `src/backends/factory.ts`. CLI commands now create the backend once and pass it as a parameter to the orchestrators instead of creating them internally.

3. **Browser Page Abstraction**:
   * Replaced Puppeteer's concrete `Page` type with a lightweight `BrowserPage` interface in all backend signatures to decouple models from specific browser libraries.

4. **Cleaned up Tool Binding**:
   * Encapsulated page tool scraping, mapping, and wrapping into a new helper class (`BrowserToolRegistry` in `src/evaluator/browser.ts`).

5. **Structured Logging & Tests**:
   * Added a configurable `ConsoleLogger` to replace direct console prints.
   * Added unit tests for the factory, registry, and logger classes. All 123 tests compile and pass.


evals-cli/src/evaluator/index.ts
```diff
@@ -8,18 +8,15 @@ import { Eval, TestResult, TestResults } from "../types/evals.js";
...
-import { OllamaBackend } from "../backends/ollama.js";
-import { VercelBackend } from "../backends/vercel.js";
 import { listToolsFromPage } from "./browser.js";
-import { SYSTEM_PROMPT } from "./prompts.js";
+import { executeInBrowserEvals } from "./browserEvaluator.js";
 
-export { listToolsFromPage };
+export { executeInBrowserEvals, listToolsFromPage };
 
 export async function executeLocalEvals(
```
swissspidy: Nit: we could also move this to a `localEvaluator.ts` file

evals-cli/src/evaluator/browser.ts
```diff
@@ -176,3 +177,37 @@ export async function listToolsFromPage(url: string): Promise<Tool[]> {
     }
   }
 }
+
+export class BrowserToolRegistry {
+  private currentTools: Tool[] = [];
+  public aiToolsWithExecution: Record<string, any> = {};
+
+  constructor(
+    private initialFallbackTools: Tool[],
```
swissspidy: Why do we have to pass an initial list of tools? It could just call `syncTools()` here.

evals-cli/src/backends/vercel.ts
```diff
@@ -114,267 +103,116 @@ export class VercelBackend implements Backend {
...
-              // Clear the object
-              for (const key in aiToolsWithExecution) {
-                delete aiToolsWithExecution[key];
-              }
-
-              // Re-populate it
-              for (const t of currentTools) {
-                aiToolsWithExecution[t.functionName] = createBrowserTool(t, page!);
-              }
+      const registry = new BrowserToolRegistry(this.tools, page);
```
swissspidy: As per my other comment, it's confusing that we have to pass a list of tools here (and to the backend constructor itself). For local evals it makes more sense, but for browser evals it's counter-intuitive.

Off the top of my head, what if we had `LocalToolRegistry` and `BrowserToolRegistry`, and we pass the registry instance to the `execute*` methods?

evals-cli/src/evaluator/index.ts
Kulikowski: Good point!

evals-cli/src/backends/vercel.ts
Kulikowski: I like this idea!

evals-cli/src/evaluator/browser.ts
Kulikowski: I'll just rewrite it into Local and Browser tools

Kulikowski:
What’s different now compared to the original PR draft:

Backend constructors are much cleaner - backends (Vercel, Gemini, Ollama) removed passing initial tools in constructors. Now that the backends pull tools dynamically from the registry at runtime.

Improved types (less `any` 👍 )

Fewer browser launches - tools are only read inside the agent loop (no initial tools as mentioned above). Since the new registry handles tool syncing dynamically - removed that pre-run call.

@swissspidy FYI

Kulikowski:
Bonus: integrated oxfmt auto-formatting into build and test scripts ;)

maudnals (approved):
LGTM, thanks for the changes

swissspidy (approved):
🚢  it

PR #299, opened 2026-07-17
Adding evals for hotel chain

Kulikowski:
Adding evals for hotel chain - commonly demoed user journeys.

Updated run_evals.sh script:

- It builds the evals-cli every time in case developer made any changes
- New target: `all` to run all the evals
- You can pass a number of desired runs for easier usage

image

### DIFF
--- a/evals-cli/run_evals.sh
+++ b/evals-cli/run_evals.sh
@@ -5,15 +5,80 @@
 # Exit immediately if a command exits with a non-zero status.
 set -e
 
-MODEL="gemini-3.5-flash"
+# Always install and build dependencies once at the start of the execution
+echo "📦 Installing dependencies and building project..."
+npm ci
+npm run build
+echo
+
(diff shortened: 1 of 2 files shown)


evals-cli/run_evals.sh
```diff
@@ -5,61 +5,99 @@
...
 set -e
 
-MODEL="gemini-3.5-flash"
+# Always install and build dependencies once at the start of the execution
+echo "📦 Installing dependencies and building project..."
+npm ci
+npm run build
+echo
+
+MODEL="google:gemini-3.1-flash-lite"
```
maudnals, line 14: Model changed, is this intentional?

maudnals (approved):
Evals LGTM! Left one question

evals-cli/run_evals.sh
Kulikowski: Yes, evals are now running faster (model is more lightweight - please let me know if that works

PR #300, opened 2026-07-20
Add timestamps to eval reports

### LINKED ISSUE #279: Enhance evals reports workflow
A common developer workflow (including when developing for evals themselves) is to launch evals, analyze the HTML evals report, make a change and then re-run the evals. 
Ideas to reduce friction when doing this repeatedly:

1. Automatically open the report in the browser once generated
2. Display the number of runs
3. Display date/timestamp (local time with explicit timezone) in the report UI and in its HTML title, instead of the static `WebMCP Evals result` to make them easier to differentiate. (The report filename already includes universal time to avoid collisions so we can display something more dev-friendly)
4. Write generated reports in a dedicated subdirectory like `evals-report` instead of root

@Kulikowski WDYT?

maudnals:
Fixes #279

Add compact timestamp to the report HTML title, and detailed one to the report UI.

HTML title:
image

Report UI:
image

### DIFF
--- a/evals-cli/src/report/report.ts
+++ b/evals-cli/src/report/report.ts
@@ -18,14 +18,33 @@ function escapeHtml(str: string | null | undefined): string {
     .replace(/'/g, "&#039;");
 }
 
+function getDetailedTimestamp(date: Date = new Date()): string {
+  return date.toLocaleString("en-US", {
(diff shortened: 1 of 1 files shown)


Kulikowski (approved):
LGTM

PR #301, opened 2026-07-20
Add skill lock configuration for evals-cli

maudnals:
Add skill lock configuration for evals-cli, and vercel skill as per https://ai-sdk.dev/.
Are we happy with this living under evals-cli considering the evals tooling needs specific dependencies not needed in other directories, or do we think managing skills globally for webmcp-tools makes more sense?

Side note: while installing this skill I saw Vercel made a skill dedicated to help us migrate the AI SDK to the next version, could come in handy

### DIFF
--- a/evals-cli/.gitignore
+++ b/evals-cli/.gitignore
@@ -3,3 +3,4 @@ dist/
 .DS_Store
 .env
 report*.html
+.agents/skills/*
--- a/evals-cli/skills-lock.json
+++ b/evals-cli/skills-lock.json
@@ -1,0 +1,11 @@
+{
+  "version": 1,
+  "skills": {
+    "ai-sdk": {
+      "source": "vercel/ai",
+      "sourceType": "github",
+      "skillPath": "skills/use-ai-sdk/SKILL.md",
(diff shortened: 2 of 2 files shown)


Kulikowski:
Looks good to me! Are you still planning more changes in this PR? I think going forward evaluating cases with SKILLS available will be very much needed

maudnals:
@Kulikowski  Thanks!

> Looks good to me! Are you still planning more changes in this PR?

No, it's more in case we need more skills here that you folks are using and I may not be aware of. If we're OK with this, happy to merge.


>  Evaluating cases with SKILLS available will be very much needed

Can you share a bit more on this? Like skills specific to testing?

Kulikowski:
SKILLS specific to journeys - my bet is that going forward we will need a way to provide SKILLS for very complex worflows, like we have 140 tools, and a skill on how to complete specific journey with 10 tools only. It is still very early and there is no standardized way of providing SKILLS for WebMCP :)

maudnals:
Definitely! Nice idea, I see how skills would make a lot of sense as journeys get more complex and involve more tools.

Changed this PR to non-draft, feel free to review and I'll merge when LGTY.

maudnals:
Closing for now, until https://github.com/vercel-labs/skills/issues/283 is merged and skills can be synced.

PR #305, opened 2026-07-20
Handle manual form submissions in evals-cli

swissspidy:
## Summary

Fixes #66

When WebMCP forms do not use the `toolautosubmit` attribute, the browser activates the form and fires a `toolactivated` DOM event rather than automatically submitting. In a headless `evals-cli` session, this previously caused the browser evaluation to hang indefinitely.

This PR addresses the issue by handling `toolactivated` events gracefully, adding tool execution result verification to evals, and reflecting result comparisons in both the console and HTML reports.

## Changes

1. **Graceful `toolactivated` Handling in `BrowserToolRegistry`**:
   - Listens for the `toolactivated` DOM event when executing browser tools.
   - If no auto-submission occurs within a 1-second timeout after `toolactivated` fires, resolves the tool execution with `"pending form submission"`.
   - Cleans up unused `listToolsFromPage` function.

2. **Tool Execution Result Matching (`result` in `expectedCall`)**:
   - Added an optional `result?: unknown;` field to `FunctionCall` and `ToolCall` types.
   - Updated `VercelBackend` to capture the tool execution return value on each executed `ToolCall`.
   - Updated `functionCallOutcome` in `utils.ts` to evaluate `expected.result` against `actual.result` using `matchesArgument` (supports exact match, objects, regex `$pattern`, `$contains`, etc.).

3. **Report and Console Enhancements**:
   - Added a "Result" comparison row to the HTML report's Step Details table.
   - Added a "Details" column to the console summary table to clearly display failure reasons (such as result mismatches).

4. **Example Evals for French Bistro Demo**:
   - Updated `evals.json` to assert `"result": "pending form submission"` when testing manual form submission behavior.
   - Added `evals-toolautosubmit.json` to assert `"result": { "$contains": "welcoming you" }` when testing auto-submission behavior.

## Examples

Expecting manual form submission:

manual-submission-success

Expected auto form submission:

toolautosubmit-success

Expecting manual form submission on a form with `toolautosubmit` results in an error:

expected-manual-submission-but-got-autosubmit

Expecting auto form submission ohn a form without `toolautosubmit` results in an error:

Screenshot 2026-07-20 at 15 36 24

### DIFF
--- a/evals-cli/src/backends/vercel.ts
+++ b/evals-cli/src/backends/vercel.ts
@@ -87,9 +87,16 @@ export class VercelBackend implements Backend {
     for (const step of aiResult.steps ?? []) {
       for (const call of (step.toolCalls ?? []) as any[]) {
         if (validToolNames.has(call.toolName)) {
+          const matchingResult: any = (step.toolResults ?? []).find(
+            (r: any) => r.toolCallId === call.toolCallId || r.toolName === call.toolName,
+          );
+          const result = matchingResult
+            ? (matchingResult.result ?? matchingResult.output)
+            : undefined;
           toolCalls.push({
             functionName: call.toolName,
             args: call.input || call.args || call.arguments || {},
+            result,
           });
         }
       }
@@ -185,13 +192,22 @@ export class VercelBackend implements Backend {
 
       // Gather executed tool calls across all steps
       const executedCalls: ToolCall[] = [];
-      if (resultPayload.steps && resultPayload.steps.length > 0) {
-        for (const step of resultPayload.steps) {
+      const stepsToIterate =
+        resultPayload.steps && resultPayload.steps.length > 0 ? resultPayload.steps : stepsHistory;
+      if (stepsToIterate && stepsToIterate.length > 0) {
+        for (const step of stepsToIterate) {
           if (step.toolCalls && step.toolCalls.length > 0) {
             for (const call of step.toolCalls) {
+              const matchingResult: any = (step.toolResults ?? []).find(
+                (r: any) => r.toolCallId === call.toolCallId || r.toolName === call.toolName,
+              );
+              const result = matchingResult
+                ? (matchingResult.result ?? matchingResult.output)
+                : undefined;
               executedCalls.push({
                 functionName: call.toolName,
                 args: (call as any).input || (call as any).args || (call as any).arguments || {},
+                result,
               });
             }
           }
--- a/evals-cli/src/backends/vercel.ts
+++ b/evals-cli/src/backends/vercel.ts
@@ -87,9 +87,14 @@ export class VercelBackend implements Backend {
     for (const step of aiResult.steps ?? []) {
       for (const call of (step.toolCalls ?? []) as any[]) {
         if (validToolNames.has(call.toolName)) {
-          const matchingResult: any = (step.toolResults ?? []).find(
-            (r: any) => r.toolCallId === call.toolCallId || r.toolName === call.toolName,
+          const matchingResult: any = (step.toolResults ?? []).find((r: any) =>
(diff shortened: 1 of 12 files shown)


evals-cli/src/evaluator/browser.ts
```diff
@@ -119,12 +73,36 @@ export class BrowserToolRegistry implements ToolRegistry {
...
-                }
+                return new Promise((resolve) => {
+                  let timer: any = null;
+
+                  const onActivated = (e: any) => {
+                    if (!e.toolName || e.toolName === name) {
+                      timer = setTimeout(() => {
+                        window.removeEventListener("toolactivated", onActivated);
+                        resolve({ success: true, data: "pending form submission" });
+                      }, 1000);
```
Kulikowski, line 84: nit: For slow tools execution we might get false "pending form submission" right? We can revisit it later if it becomes a problem (maybe add some logger in case timeout happens so we know about it?)

evals-cli/src/backends/vercel.ts
```diff
@@ -87,9 +87,16 @@ export class VercelBackend implements Backend {
     for (const step of aiResult.steps ?? []) {
       for (const call of (step.toolCalls ?? []) as any[]) {
         if (validToolNames.has(call.toolName)) {
+          const matchingResult: any = (step.toolResults ?? []).find(
```
Kulikowski: nit: If tool-call-id doesn't match we are falling back to name check, ; We might want to log not-matching-call-id for cases where same tool is called multiple times (like "load-next-results")

Kulikowski: Do we need name-matching?

evals-cli/src/evaluator/browser.ts
Kulikowski: It would be cool to check if tool is form based and has no toolautosubmit 🤔

Kulikowski:
+ I think we should add unit tests here

Kulikowski (changes requested):

evals-cli/src/evaluator/browser.ts
swissspidy: Yeah I need to double check this with the browser implementation. The event might not even get fired for autosubmit, in which case the 1s timeout could be removed.

beaufortfrancois: FYI Puppeteer allows you to see if the tool is a declarative tool. See https://pptr.dev/guides/webmcp#declarative-registration 

The toolactivated event is currently always fired regardless of toolautosubmit. See https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/script_tools/model_context.cc;l=555;drc=56e29035203bd490455a5ca1cd1207f782c3442b

swissspidy: Ooh, very helpful, thanks!

swissspidy: I realized we don't actually use Puppeteer's WebMCP API yet (we should probably do that in another PR.). And from JS alone we can't see whether it's declarative or not, AFAICT.

beaufortfrancois: Instead of `window.ontoolactivated`, use the following: 
```js
page.webmcp.on('toolinvoked', call => {
  console.log(`Tool ${call.tool.name} was invoked with input:`, call.input);
});

evals-cli/src/backends/vercel.ts
swissspidy: Logging added.

evals-cli/src/evaluator/browser.ts
swissspidy: Working on this in #326

PR #308, opened 2026-07-21
Implement analyzer for evals report

### LINKED ISSUE #302: Add evals report analyzer
As discussed with team (@Kulikowski , @swissspidy): add a feature to help developers make sense of the evals reports, by summarizing and analyze the results (single report, or compare multiple across subsequent evals runs):
* As a parameter for a given evals run
* As a standalone command to analyze 1-n evals report(s) by specifying the file path

maudnals:
Fixes #302  

* Eval report analysis is available as a standalone command or as a flag. To test with the sports shop:
```
node dist/bin/webmcp-evals.js browser \
  -u https://googlechromelabs.github.io/webmcp-tools/demos/sport-shop-angular/ \
  -e examples/sport-shop-angular/evals.json \
  --analyze --open --open-analysis
```
* I did include an optimization (see analyzer index.ts l.140) since it offers big token and time savings (side note, it seems the HTML report is smaller in tokens because it only displays the tool name and its high-level description vs the JSON report. The JSON report is probably more suited for the analyzer since it's more structured. Something to explore a bit further, though.)
* I introduced new flags to keep full flexibility / decoupling between main eval reports generation and analysis report generation: `--analyzer-model` and `--open-analysis`. LMK if there's a nicer way!
* I've tested the analyzer on the sports shop. Results are OK but need refinement. Fixing this requires us to take a closer look on a case by case basis, label as humans whether we think the analyzer did the right thing, and tweak its prompt accordingly. My suggestion would be we keep this PR focused on wiring the feature together, and addressing quality improvements in a follow-up. WDYT?

Here's one example where IMO, the analyzer is wrong, but let me know what you think:
On sports shop, the analyzer says: 
> _The phrase "Apply filter for 50$" contains linguistic ambiguity. The model mapped the literal token "50$" to`"50-99.99"` bin rather than recognizing that "under $50" (indicated in the test name and implicit in youth gift requests) mapped more appropriately to `"0-49.99"`._


swissspidy:
Great idea to use Modern Web Guidance for this! 👍

maudnals:
> Great idea to use Modern Web Guidance for this! 👍

I think so too! We initially prototyped this with a static set of WebMCP resources, but @Kulikowski suggested MWG, which is a win-win for both projects.

Kulikowski (approved):
LGTM

maudnals:
Added unit tests!

PR #309, opened 2026-07-21
Code mode follow-up changes

swissspidy:
Addressing some code review comments from #287


demos/page-agent/script.js
```diff
@@ -95,6 +92,7 @@ function getTSType(schema) {
 }
 
 async function executeBatchLocally(steps) {
+  const { executeDeclarativeBatch } = await import('../shared/webmcp-batch.js');
```
beaufortfrancois: You may want to save this promise so that subsequent calls can use `executeDeclarativeBatch` directly. WDYT?

demos/shared/webmcp-batch.js
```diff
@@ -132,7 +132,7 @@ export function registerExecuteBatchTool(modelContext = typeof window !== 'undef
         outputs
       };
     }
-  }, { exposedTo: ['*'] }).catch(() => {});
```
beaufortfrancois, line 135: How was this working for you before then?

swissspidy: Good question 😄  maybe because I was testing on localhost, or Chrome Stable & the polyfill kicked in or so 🤷

demos/page-agent/script.js
swissspidy: Done ✅  I (wrongly?) thought subsequent imports would be cached by the browser anyway.

beaufortfrancois: Sorry! I've checked locally and it seems like subsequent imports resolve instantly without going through DevTools network.

swissspidy: Thanks for testing! Let me revert this then :)

demos/page-agent/script.js
```diff
@@ -22,11 +21,9 @@ let ai, chat;
 
 codeModeCheckbox.addEventListener('change', () => {
   chat = null;
-  chatWindow.innerHTML = '';
   appendMessage('System', `🔄 Switched to ${codeModeCheckbox.checked ? 'Code Mode' : 'Normal Mode'}. Chat restarted.`, 'tool-indicator');
 });
```
beaufortfrancois, line 48: It would be nice to see how many tools are now exposed by the iframe. There should be one more. Executing same code as iframe.onload listener should be done after `Switched to ...`.

demos/page-agent/script.js
beaufortfrancois: Since `webmcp-batch.js` also injects register execute_batch tool, we should make sure:
1. execute_batch tool is registered when code mode checkbox is checked. 
2. execute_batch tool is actually registered before executing it.
3. execute_batch tool is unregistered when code mode checkbox is disabled

beaufortfrancois (changes requested):

demos/page-agent/script.js
```diff
@@ -19,14 +18,46 @@ const iframe = document.getElementById('iframe');
...
+    if (!codeModeAbortController) {
+      codeModeAbortController = new AbortController();
+      const { registerExecuteBatchTool } = await import('../shared/webmcp-batch.js');
+      await registerExecuteBatchTool(document.modelContext, { signal: codeModeAbortController.signal });
+    }
+  } else {
+    if (codeModeAbortController) {
+      codeModeAbortController.abort();
+      codeModeAbortController = null;
+      document.modelContext.unregisterTool?.('execute_batch');
```
beaufortfrancois: `unregisterTool` doesn't exist anymore.

demos/shared/webmcp-batch.js
```diff
@@ -80,10 +80,10 @@ export async function executeDeclarativeBatch(steps, executeToolFn) {
   return outputs;
 }
 
-export function registerExecuteBatchTool(modelContext = typeof window !== 'undefined' ? window.document?.modelContext : undefined) {
+export function registerExecuteBatchTool(modelContext = typeof window !== 'undefined' ? window.document?.modelContext : undefined, options) {
```
beaufortfrancois: How about we remove modelContext param for now and only add it back when we actually need it? 
We should use `return document.modelContext.registerTool({...` only

demos/page-agent/script.js
```diff
@@ -228,18 +217,14 @@ function loadUrl() {
...
-    'tool-indicator',
-  );
-});
+iframe.addEventListener('load', logExposedTools);
 
-logoutBtn.addEventListener('click', () => {
+logoutBtn.addEventListener('click', async () => {
   localStorage.removeItem('gemini_api_key');
   ai = null;
+  chat = null;
```
beaufortfrancois: This is not needed since handleUserSubmit is gated by `ai`.

demos/page-agent/script.js
```diff
@@ -262,6 +247,8 @@ async function handleUserSubmit() {
 
     appendMessage('You', text, 'user');
 
+    await syncCodeModeTool();
```
beaufortfrancois: Why is this needed?

demos/page-agent/script.js
```diff
@@ -297,31 +284,26 @@ async function handleUserSubmit() {
...
             const tools = await getTools();
             const tool = tools.find((t) => t.name == name);
-            
-            let result;
-            if (name === 'execute_batch' && !tool) {
-              result = await executeBatchLocally(args.steps);
-            } else {
-              result = await document.modelContext.executeTool(tool, inputArgs);
+            if (!tool) {
+              throw new Error(`Tool ${name} not found`);
```
beaufortfrancois: This should be covered already by `document.modelContext.executeTool` since it will reject?

demos/page-agent/script.js
```diff
@@ -297,31 +284,26 @@ async function handleUserSubmit() {
             appendMessage('System', `⚙️ Executing tool: ${name}...`, 'tool-indicator');
             const tools = await getTools();
             const tool = tools.find((t) => t.name == name);
-            
-            let result;
-            if (name === 'execute_batch' && !tool) {
-              result = await executeBatchLocally(args.steps);
```
beaufortfrancois, line 303: Removing this makes things smoother eventually. Nice!

demos/page-agent/script.js
```diff
@@ -297,31 +284,26 @@ async function handleUserSubmit() {
...
-                    appendMessage(
-                      'Console',
-                      `❌ Step [${out.id || 'anonymous'}]: call to <strong>${out.tool}</strong> failed.<br>↳ Error: <span style="color:red">${out.error}</span>`,
-                      'console-log'
-                    );
-                  }
+            
+            const result = await document.modelContext.executeTool(tool, inputArgs);
+
+            if (name === 'execute_batch' && result && Array.isArray(result.outputs)) {
```
beaufortfrancois: Shall we also check for `codeModeCheckbox.checked` as well?

beaufortfrancois (changes requested):

demos/page-agent/script.js
swissspidy: The `syncCodeModeTool` was related to https://github.com/GoogleChromeLabs/webmcp-tools/pull/309#discussion_r3622250888 to ensure the tool is registered and unregistered correctly.

demos/page-agent/script.js
```diff
@@ -94,24 +124,6 @@ function getTSType(schema) {
...
-  const outputs = await executeDeclarativeBatch(steps, executeToolFn);
-  const success = outputs.every(o => o.success);
-  return {
-    success,
-    outputs
-  };
-}
-
 async function getConfig() {
   const tools = await getTools();
```
beaufortfrancois, line 58: Nit: Remove extra spaces.

demos/page-agent/script.js
```diff
@@ -128,67 +140,43 @@ async function getConfig() {
...
   const systemInstruction = [
     'You are an assistant embedded in a web page.',
     'CRITICAL RULE: Do not try to use other tools than the available ones.',
-  ].join('\n');
+  ];
 
-  const functionDeclarations = tools.map((tool) => {
-    return {
+  const functionDeclarations = tools
+    .filter((tool) => tool.name !== 'execute_batch')
```
beaufortfrancois: My understanding is that `execute_batch` shouldn't be there in normal mode so we don't need to exclude it right?

demos/page-agent/script.js
```diff
@@ -128,67 +140,43 @@ async function getConfig() {
...
+      .filter((tool) => tool.name === 'execute_batch')
+      .map((tool) => ({
+        name: tool.name,
+        description: tool.description,
+        parametersJsonSchema: tool.inputSchema
+          ? JSON.parse(tool.inputSchema)
+          : { type: 'object', properties: {} },
+      }));
 
     return {
```
beaufortfrancois: Formatting nit: `return { systemInstruction, tools: [{ functionDeclarations }] };`

demos/page-agent/script.js
beaufortfrancois: Please run default prettier formatter

demos/page-agent/script.js
```diff
@@ -19,14 +18,45 @@ const iframe = document.getElementById('iframe');
 const codeModeCheckbox = document.getElementById('code-mode-checkbox');
 
 let ai, chat;
+let codeModeAbortController = null;
 
-codeModeCheckbox.addEventListener('change', () => {
+async function syncCodeModeTool() {
+  if (!document.modelContext) return;
```
beaufortfrancois: Nit: Remove this line.

demos/page-agent/script.js
```diff
@@ -19,14 +18,45 @@ const iframe = document.getElementById('iframe');
...
+
+  if (codeModeCheckbox.checked) {
```
beaufortfrancois: I think you can now simply move this code into codeModeCheckbox change listener:

```js
codeModeCheckbox.addEventListener('change', async () => {
  chat = null;

  codeModeAbortController?.abort();
  if (codeModeCheckbox.checked) {
    codeModeAbortController = new AbortController();
    const { registerExecuteBatchTool } = await import('../shared/webmcp-batch.js');
    await registerExecuteBatchTool({ signal: codeModeAbortController.signal });
  }

  appendMessage(
    'System',
    `🔄 Switched to ${codeModeCheckbox.checked ? 'Code Mode' : 'Normal Mode'}. Chat restarted.`,
    'tool-indicator',
  );
  await logExposedTools();
});
```

demos/page-agent/script.js
beaufortfrancois: Those seem very specific to webmcp-batch.js? Should we have a `getSystemInstructions(tools)` function in this shared file so that we can move generateTSDeclaration and friends into it?

demos/page-agent/script.js
```diff
@@ -128,67 +140,43 @@ async function getConfig() {
...
-    };
+    ];
+
+    const functionDeclarations = tools
+      .filter((tool) => tool.name === 'execute_batch')
+      .map((tool) => ({
+        name: tool.name,
+        description: tool.description,
+        parametersJsonSchema: tool.inputSchema
+          ? JSON.parse(tool.inputSchema)
```
beaufortfrancois: Since we do inject `execute_batch`, we don't need to check whether inputSchema is there. We know it is. WDYT?

beaufortfrancois (changes requested):
I think we're almost there!

demos/page-agent/script.js
swissspidy: What extra space? This line is unmodified in this PR.

There's no formatter configuration set up in the project. evals-cli/ has `oxfmt` but that's about it.

I suggest keeping formatting changes to another PR.

demos/page-agent/script.js
swissspidy: As per https://github.com/GoogleChromeLabs/webmcp-tools/pull/309#discussion_r3628759253 let's leave formatting for another PR. It should be automated so we don't have to discuss it.

demos/page-agent/script.js
swissspidy: Yeah I'll need to revisit this.

The original idea with the webmcp-batch script is that the iframed page _or_ page-agent could provide it (or both), hence this check to avoid duplicates. But right now it's only loaded by page-agent.

demos/page-agent/script.js
swissspidy: Yeah can probably extract this.

Need to revisit this anyway, as `generateTSDeclaration` is a bit too much. Useful for actual code mode, but here for this declarative approach, just printing the schema would make more sense.

beaufortfrancois: I was thinking we could have something like:

```js
async function getConfig() {

  const systemInstruction = await getSystemInstruction(tools, codeModeCheckbox.checked);
    
  const functionDeclarations = tools
    .filter((t) => (t.name === 'execute_batch') === codeModeCheckbox.checked);
    .map((tool) => ({
      name: tool.name,
      description: tool.description,
      parametersJsonSchema: tool.inputSchema
        ? JSON.parse(tool.inputSchema)
        : { type: 'object', properties: {} },
    }));

  return { systemInstruction, tools: [{ functionDeclarations }] };
}
```

demos/shared/webmcp-batch.js
```diff
@@ -132,10 +130,41 @@ export function registerExecuteBatchTool(modelContext = typeof window !== 'undef
...
       };
     }
-  }, { exposedTo: ['*'] }).catch(() => {});
+  }, options);
 }
 
-// Auto-register in browser environment if modelContext is available
-if (typeof window !== 'undefined' && window.document && window.document.modelContext) {
-  registerExecuteBatchTool(window.document.modelContext);
+export function getSystemInstructions(tools) {
```
beaufortfrancois: Nit: Get rid of `getSystemInstructions`
```suggestion
export function getSystemInstruction(tools) {
```

beaufortfrancois (approved):
LGTM with nit

beaufortfrancois:
Yeah! FYI I've updated https://chrome.dev/web-ai-demos/webmcp/page-agent/ as well

PR #319, opened 2026-07-28
Add Dependabot configuration file

swissspidy:
This should make it easier to keep everything up-to-date across all the subfolders


beaufortfrancois (approved):
Thanks for doing this, it should improve our lives! Note that I'm no expert in Dependabot configuration files but LGTM nonetheless ;)

PR #323, opened 2026-07-30
Add new library entry for @web-ai-sdk/webmcp

obetomuniz:
Adds `@web-ai-sdk/webmcp` to the Libraries & Tools section.

It provides a zero-dependency, framework-agnostic TypeScript API for registering WebMCP tools, plus an optional React adapter.


beaufortfrancois:
@andreban @Kulikowski I'm afraid https://web-ai-sdk.dev/docs/react/use-web-mcp/ will create some confusion with https://github.com/GoogleChromeLabs/use-webmcp-tool. WDYT?

beaufortfrancois:
@obetomuniz Would you be open to contribute to https://github.com/GoogleChromeLabs/use-webmcp-tool instead of having to maintain separate react hooks for WebMCP?

obetomuniz:
> @obetomuniz Would you be open to contribute to https://github.com/GoogleChromeLabs/use-webmcp-tool instead of having to maintain separate react hooks for WebMCP? 

Thanks for the suggestion! I'm open to collaborating and aligning where it makes sense. However, my implementation predates `use-webmcp-tool`, and the React hook is only an optional adapter within the broader, framework-agnostic web-ai-sdk project. Given that wider scope, I'd prefer to continue maintaining it there as-is.

beaufortfrancois:
@Kulikowski I've reordered the "Libraries & Tools" list and added "(by Google)" mention to make it clearer which are handled by Google. If that looks good to you, I'll merge.

Kulikowski:
> @Kulikowski I've reordered the "Libraries & Tools" list and added "(by Google)" mention to make it clearer which are handled by Google. If that looks good to you, I'll merge.

LGTM

beaufortfrancois:
Merged

PR #324, opened 2026-07-31
feat(evals): add deterministic smoke runner

### LINKED ISSUE #313: Introduce non-LLM smoke-test runner based on eval files to verify demos
Currently, testing our demo user journeys relies heavily on full evaluation runs or manual testing. To catch regressions early and ensure we don't break existing demos, we need a lightweight, deterministic verification mechanism that runs end-to-end user journeys without the cost, latency, or non-determinism of an LLM.

Proposed Solution
Introduce an automated smoke-test runner that reuses our existing eval files (similar to the patterns seen in [PR #299](https://github.com/GoogleChromeLabs/webmcp-tools/pull/299/changes)).

Instead of passing prompts to an LLM to decide the next actions, the test runner will programmatically and deterministically trigger the tool calls defined in the eval files against a target website. This allows us to verify that the core journeys, tool integrations, and UI states function as expected.

Scope
Execute predefined tool calls sequence-by-sequence from eval files without LLM involvement.

Provide a fast feedback loop in CI/local development to confirm demos and core journeys are operational.

Leverage existing eval file formats so we don't need to maintain a separate set of test definitions.

Challenges & Considerations

- The runner requires an active, running web application to test against. I am considering supporting two execution modes:

- The smoke test runner automatically spins up the local app/server before executing tests and tears it down afterward.

- The runner accepts a target URL (e.g., staging or a live deployed demo) to run the checks against.

- Because these tests interact with running websites and real browser/network states, they may be prone to environmental flakiness. We should implement sensible retry logic, explicit wait states, or timeout handling to mitigate false positives.

Acceptance Criteria
- A test runner script/utility is implemented that can parse existing eval files and execute tool sequences deterministically.

- Support for configuring the target environment (e.g., via environment variables or CLI flags for local vs. deployed URLs).

- Basic error reporting that clearly highlights which tool call or step failed during the journey.

- Documentation updated with instructions on how to run the smoke tests locally and how to format eval files for compatibility.

- (Optional) Initial integration into our CI/CD pipeline to automatically run against pull requests.

alectimison-maker:
## Summary

- add `webmcp-evals smoke` to execute concrete `expectedCall` trajectories against a live WebMCP page without an LLM or API key
- report the exact case, step, and tool for missing tools, execution failures, explicit unsuccessful results, and timeouts
- validate the complete eval file before execution, isolate cases in fresh pages, and document supported smoke-test inputs

Optional calls are skipped, nested groups use authored order, and matcher constraints are rejected because guessing values would make the runner non-deterministic.

Closes #313.

## Testing

- `npm run build`
- `node --test dist/test/smokeEvaluator.test.js dist/test/browserToolRegistry.test.js` (16/16 passed)
- all non-Chrome-path tests (141/141 passed)
- `npm run lint` (0 errors; 4 pre-existing warnings)
- `npx oxfmt --check`
- `npm pack` archive inspection
- live run against the upstream pizza-maker demo using the existing eval file (9/9 passed)

The full `npm test` result was 142/143 locally. Its sole failure is the pre-existing `findChromePath` environment test because Chrome Canary is not installed at one of the repository's hard-coded paths; the live smoke run used installed Chrome through the runner's browser dependency seam.

### DIFF
--- a/evals-cli/README.md
+++ b/evals-cli/README.md
@@ -7,10 +7,11 @@ A TypeScript evaluation framework and CLI for testing the tool-calling capabilit
 
 ## Features
 
-- **CLI Interface**: Built with `commander` providing `local` and `browser` commands.
+- **CLI Interface**: Built with `commander` providing `local`, `browser`, and `smoke` commands.
 - **Execution Modes**:
   - **`local`**: Runs evaluations against static JSON tool schema definition files.
   - **`browser`**: Runs live evaluations against WebMCP tools exposed on web pages via Puppeteer.
+  - **`smoke`**: Executes concrete expected tool calls against a live page without an LLM or API key.
 - **Model Backends**: Supports `@google/genai` (`gemini`), Ollama (`ollama`), and Vercel AI SDK (`vercel`).
 - **Reporters**: Supports `console`, `json`, and `html` output to the `.evals` directory.
 - **Constraint-Based Matching**: Matches expected tool calls using regex patterns, numerical ranges, type checks, and orderings (`ordered` and `unordered`).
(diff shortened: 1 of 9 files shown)


alectimison-maker:
@googlebot rescan

Kulikowski (approved):
Thanks for your contribution! It's great!

I've added some small fixes and run_smoke.sh file to execute all the smoke tests.

PR #326, opened 2026-07-31
Evals CLI: Leverage Puppeteer WebMCP API

swissspidy:
Resolves #325

Addresses feedback on #305 regarding detecting declarative tools.

Successfully tested with the French Bistro example from #305. Examples:

```bash
# Success cases
$ node dist/bin/webmcp-evals.js browser -u https://googlechromelabs.github.io/webmcp-tools/demos/french-bistro/-e examples/french-bistro/evals.json --open # pending form submission
$ node dist/bin/webmcp-evals.js browser -u https://googlechromelabs.github.io/webmcp-tools/demos/french-bistro/?toolautosubmit -e examples/french-bistro/evals-toolautosubmit.json --open # autosubmit

# Fail cases
$ node dist/bin/webmcp-evals.js browser -u https://googlechromelabs.github.io/webmcp-tools/demos/french-bistro/ -e examples/french-bistro/evals-toolautosubmit.json --open # expected autosubmit, got pending form submission
$ node dist/bin/webmcp-evals.js browser -u https://googlechromelabs.github.io/webmcp-tools/demos/french-bistro/?toolautosubmit -e examples/french-bistro/evals.json --open  # expected pending form submission, got autosubmit
```

### Summary of Changes

1. **Leveraging `page.webmcp`**:
   - Updated `PUPPETEER_FLAGS` with `--enable-features=WebMCPTesting,DevToolsWebMCPSupport,WebMCP`.
   - Updated `getToolsFromBrowserPage` to use `page.webmcp.tools()`.
   - Updated `BrowserToolRegistry.executeTool` to use `page.webmcp.tools()` and `tool.execute()`.
2. **Declarative Form Tools**:
   - Handled non-autosubmit declarative form tools (`tool.formElement` && `!tool.annotations?.autosubmit`) via `page.webmcp.on('toolinvoked')` event listener.
3. **Tests**:
   - Updated `BrowserPage` type and test mocks in `vercel.test.ts`.
   - Added unit tests in `browserToolRegistry.test.ts` for native WebMCP tool discovery and execution.

PR #340, opened 2026-07-31
Remove json parsing in mapRawBrowserToolsToConfig

swissspidy:
Follow-up to #326, see https://github.com/GoogleChromeLabs/webmcp-tools/pull/326#discussion_r3690972783


evals-cli/src/test/mappers.test.ts
beaufortfrancois, line 180: You should try no inputSchema

beaufortfrancois (approved):
LGTM with nit

PR #341, opened 2026-08-01
feat: enhance evals CLI logs with progress bar and detailed table structure

### LINKED ISSUE #282: [Feature Request] Improve CLI logs: add progress bar and update table
As discussed with @Kulikowski:
* A basic version would be to do doneSteps / totalStepCount. Each test case can have 1 step (tool call) - n steps (= tool calls) 
* Fix the in-CLI step preview at the moment which displays a wrong step count (remember step count is unknown ahead of time!)
*  Group the final CLI summary table by Test Case -> Run -> Step, for consistency with the reports

maudnals:
Fixes #282 
* Added a progress bar. Progress is calculated as test cases done / total test cases (= test case per run * run count). I chose this approach because it's step-agnostic.
* Grouped the final CLI summary table by Test Case -> Run -> Step, for consistency with the HTML report and more clarity on the run and test cases
* I left the final pass count as step-based for now, event though I'm not sure that's the most intuitive, but that's a broader design decision
* I also removed dotenv logs for less noise, let me know if there's a reason we should keep them by default

image

image

image

### DIFF
--- a/evals-cli/src/backends/vercel.ts
+++ b/evals-cli/src/backends/vercel.ts
@@ -87,9 +87,21 @@ export class VercelBackend implements Backend {
     for (const step of aiResult.steps ?? []) {
       for (const call of (step.toolCalls ?? []) as any[]) {
         if (validToolNames.has(call.toolName)) {
+          const matchingResult: any = (step.toolResults ?? []).find((r: any) =>
+            call.toolCallId ? r.toolCallId === call.toolCallId : r.toolName === call.toolName,
+          );
+          if (!matchingResult && call.toolCallId) {
+            this.logger.debug(
+              `[DEBUG] Could not find matching tool result for call ID "${call.toolCallId}" (${call.toolName})`,
+            );
+          }
+          const result = matchingResult
+            ? (matchingResult.result ?? matchingResult.output)
+            : undefined;
           toolCalls.push({
             functionName: call.toolName,
             args: call.input || call.args || call.arguments || {},
+            result,
           });
         }
       }
@@ -118,7 +130,7 @@ export class VercelBackend implements Backend {
 
     try {
       const registry = new BrowserToolRegistry(page);
-      let currentTools = await registry.syncTools();
+      let currentTools = registry.syncTools();
 
       if (currentTools.length === 0) {
         const executablePath = await findChromePath();
@@ -173,7 +185,7 @@ export class VercelBackend implements Backend {
           });
         },
         prepareStep: async (_opts: any): Promise<any> => {
-          currentTools = await registry.syncTools();
+          currentTools = registry.syncTools();
           rebuildTools(currentTools);
           availableToolsPerStep.push([...currentTools]);
           return _opts;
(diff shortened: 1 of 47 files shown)


evals-cli/src/commands/index.ts
```diff
@@ -162,27 +170,38 @@ export async function runWebCommand(options: CommandOptions, command?: Command):
...
+          );
         } else if (event.type === "progress") {
           resultsList.push(event.result);
-          const passRate = (
-            (resultsList.filter((r) => r.outcome === "pass").length / resultsList.length) *
-            100
-          ).toFixed(2);
-          spinner.text = `Running... pass rate: ${passRate}% (${resultsList.length} steps)`;
+          const currentRun = event.result.runIndex || 1;
+          const caseIndex = event.testNumber - (currentRun - 1) * tests.length;
```
Kulikowski: To verify: When we run multiple multi step evals is the progress bar loading properly?

evals-cli/src/commands/index.ts
```diff
@@ -162,27 +170,38 @@ export async function runWebCommand(options: CommandOptions, command?: Command):
     const resultsList: any[] = [];
 
     if (useConsole) {
+      console.log();
       spinner = ora({ discardStdin: false });
     }
 
     const backend = createBackend(config);
     const finalResults = await executeInBrowserEvals(tests, backend, config, (event) => {
```
Kulikowski, line 178: I think we can do the same for Local Evals run (this is for Browser one)

evals-cli/src/commands/index.ts
```diff
@@ -122,6 +122,14 @@ export async function runLocalCommand(options: CommandOptions, command?: Command
   }
 }
 
+function getProgressBar(ratio: number, size = 10): string {
```
Kulikowski: To check: RangeError?

Kulikowski (approved):
LGTM with comments

Kulikowski:
+ I think we should add unit tests :)

evals-cli/src/commands/index.ts
maudnals: Yes, the progress bar is decoupled from the step count (tool calls/assertions count), it's based on the number of test cases

evals-cli/src/commands/index.ts
maudnals: SG, thanks for catching this! Done.

evals-cli/src/commands/index.ts
maudnals: Nice catch, added safeguard

maudnals:
> * I think we should add unit tests :)

Added :)

evals-cli/src/commands/index.ts
```diff
@@ -122,6 +115,16 @@ export async function runLocalCommand(options: CommandOptions, command?: Command
   }
 }
 
+export function getProgressBar(ratio: number, size = 10): string {
```
9kubczas4, line 123: nit: maybe we can move this function to separate file like `utils/console.ts`

evals-cli/src/backends/vercel.ts
9kubczas4, line 204: Idea for next PR: from my point of view code to calculate executed calls looks quite similar inside `executeInBrowserEval` and `executeLocalEvals`. Maybe we could share the logic.

evals-cli/src/evaluator/mappers.ts
9kubczas4, line 99:
```
  return rawTools.map((t) => ({
    description: t.description,
    functionName: t.name,
    parameters: t.inputSchema || {},
  }));
```

Kulikowski:
LGTM

evals-cli/src/commands/index.ts
maudnals: Let's keep this PR's scope focused, but thanks for the suggestion, good idea! Feel free to open a ticket and separate PR for the utils reorganization if you'd like.

evals-cli/src/backends/vercel.ts
maudnals: Good point, and as you noted this is out of scope for this branch. Feel free to open an issue to track and discuss this, contributions are welcome :)

PR #344, opened 2026-08-03
simplify tool parameter descriptions for reservation date and time

beaufortfrancois:
As noted in https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/html/forms/form_mcp_schema.cc;l=583;drc=519653dfe228949fce89c3cf8f7a6152f5531fcb, reservation date toolparamdescription doesn't need to include spefically `(YYYY-MM-DD)`.
I've tried without `(HH:MM)` and it worked great as well with Gemini 3.5 Flash Lite and Gemini 3.6

### DIFF
--- a/demos/french-bistro/index.html
+++ b/demos/french-bistro/index.html
@@ -77,7 +77,7 @@
               id="date"
               name="date"
               required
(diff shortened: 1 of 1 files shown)


Kulikowski (approved):
LGTM

PR #348, opened 2026-08-04
feat: add --chrome-channel CLI option to specify browser channel

### LINKED ISSUE #343: Add possibility to define channel (Canary, Stable etc) to run evals with
Currently evals are run with Chrome Canary - we should provide a way to configure it via `--channel` flag

beaufortfrancois:
FIX https://github.com/GoogleChromeLabs/webmcp-tools/issues/343

### DIFF
--- a/evals-cli/README.md
+++ b/evals-cli/README.md
@@ -70,16 +70,17 @@ src/
 
 Shared across commands:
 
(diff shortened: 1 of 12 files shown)


Kulikowski (approved):
LGTM

PR #353, opened 2026-08-05
fix(evals): surface unexpected tool calls in reports

### LINKED ISSUE #283: [Feature Request] Unexpected tool calls should be surfaced
Due to model hallucination we can see attempts to call tools that are non-existent. 

We should properly mark it in the report to make sure that it's visible.

Currently when we have more tool calls than expected, the matcher will match it as "null" expected pattern which can be confusing.

alectimison-maker:
## Summary

- label extra model executions as `Unexpected tool call` in HTML eval reports
- replace the blank expected-function cell with `No call expected`
- add regression coverage that distinguishes an unexpected call from a missing expected call

## Motivation

The trajectory matcher already fails executions that have no matching expected call, but the HTML report renders an empty expected pattern for those rows. This makes hallucinated or otherwise extra tool calls difficult to identify when reviewing an eval report.

## Design

The existing `TestResult` shape already represents this case unambiguously: it has an actual response and no expected call. The report renderer now detects that combination and adds a focused failure notice while leaving matching, evaluator behavior, JSON output, and report grouping unchanged.

This keeps the presentation fix local to the HTML renderer instead of adding a new reason field to shared trajectory results or trying to infer whether the call was absent from a particular tool registry.

## Testing

- `npm test` — 149 passed, 1 skipped, 0 failed (the browser integration test skipped because no configured browser could be launched)
- `node --test dist/test/report.test.js` — 4 passed
- `npm run build` — passed
- `npm run lint` — passed with 7 pre-existing warnings in untouched files
- `npx oxfmt --check` — passed

## Compatibility and risks

There are no changes to matching semantics, exported types, CLI flags, JSON reports, or persisted formats. The actual function name is escaped before it is included in the HTML notice. The change applies to both local and browser HTML reports through their shared step renderer.

## Scope

This change only clarifies extra-call failures. Broader report redesign and tool-registry-based unknown-tool classification are intentionally out of scope.

Closes #283

### DIFF
--- a/evals-cli/src/report/report.ts
+++ b/evals-cli/src/report/report.ts
@@ -350,6 +350,7 @@ function renderRunIteration(run: TestRun, totalRuns: number): string {
 function renderStepDetails(stepEval: TestStep, totalSteps: number): string {
   const { stepIndex, result } = stepEval;
   const expected = result.test.expectedCall?.[0] as FunctionCall | undefined;
+  const isUnexpectedCall = expected === undefined && result.response !== null;
 
   const functionNameOutcome =
     expected?.functionName === result.response?.functionName ? "pass" : "fail";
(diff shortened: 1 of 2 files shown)


maudnals (approved):
LGTM

PR #361, opened 2026-08-06
feat(evals): include browser console errors in reports

alectimison-maker:
## Summary

- capture browser `console.error` messages and uncaught page exceptions while WebMCP tools execute
- attribute each diagnostic to the active tool name and arguments, including all possible sources when tool calls overlap
- show diagnostics once per run in HTML reports, with a visible error-count badge and source location when available

## Motivation

Browser evals can trigger JavaScript failures while an agent executes WebMCP tools. Those failures are visible in DevTools, but the generated eval report currently drops them, which makes a broken tool implementation and an incorrect model trajectory harder to distinguish.

Collecting every page-level error would also include unrelated startup and background noise. This update therefore records errors only inside WebMCP tool execution and includes the relevant tool context in the report.

## Design

`BrowserToolRegistry` installs Puppeteer `console` and `pageerror` listeners when the first tool call starts and removes them after the last active call finishes. Only error-level console messages and uncaught exceptions are collected; ordinary logs, warnings, startup errors, and post-call background errors are ignored.

Each diagnostic snapshots the active tool names and arguments. A single active call is shown as the source. If calls overlap, the report lists every active call as a possible source rather than making a false causal claim or duplicating the error.

Diagnostics are returned through an optional `browserConsoleErrors` result field. A multi-step run attaches the array to its first result only, avoiding repeated data in JSON reports. The HTML renderer groups that field at the run level, escapes page- and model-controlled content, and opens runs with browser errors while preserving the existing eval outcome.

## Testing

- `npm test` — 153 passed, 1 skipped, 0 failed (the existing browser integration test skipped because no configured Chrome channel could be launched locally)
- `npm run build && node --test dist/test/browserConsole.test.js dist/test/report.test.js dist/test/browserToolRegistry.test.js dist/test/vercel.test.js` — 21 passed
- `npm run lint` — passed with 7 pre-existing warnings in untouched files
- `npx oxfmt --check` — passed

## Compatibility and risks

This does not change trajectory matching, pass/fail counts, CLI flags, or local eval behavior. The JSON result shape gains an optional field that is omitted when a browser run has no attributed errors.

Console errors and tool arguments can contain application data, so generated reports should be handled with the same care as the prompts, tool arguments, and results they already contain. Tool context, messages, and source URLs are HTML-escaped before rendering.

When tool executions genuinely overlap, browser events cannot identify which active call emitted an error. The report explicitly lists all active calls as possible sources.

## Scope

This change intentionally does not capture normal logs, warnings, initial page-load errors, or background errors outside tool execution. Browser console errors do not automatically fail an otherwise valid eval. Configurable console levels and failure policies are left for separate changes.

Closes #62

### DIFF
--- a/evals-cli/src/evaluator/browserEvaluator.ts
+++ b/evals-cli/src/evaluator/browserEvaluator.ts
@@ -89,9 +89,9 @@ async function runSingleBrowserTest(
   let browserConsoleErrors: BrowserConsoleError[] = [];
   try {
     page = await browser.newPage();
-    browserConsoleErrors = collectBrowserConsoleErrors(page);
     await setupBrowserPage(page, config.url);
     const evalResult = await backend.executeInBrowserEval(test, page, config);
+    browserConsoleErrors = evalResult.browserConsoleErrors || [];
 
     if (evalResult.error) {
       throw evalResult.error;
@@ -129,33 +129,6 @@ async function setupBrowserPage(page: Page, url: string): Promise<void> {
     waitUntil: "networkidle2",
     timeout: 30000,
   });
-}
-
-export function collectBrowserConsoleErrors(page: Page): BrowserConsoleError[] {
-  const errors: BrowserConsoleError[] = [];
-
-  page.on("console", (entry) => {
-    if (entry.type() !== "error") return;
-
-    const location = entry.location();
-    const error: BrowserConsoleError = {
-      kind: "console",
-      message: entry.text(),
-    };
-    if (location.url) error.url = location.url;
-    if (location.lineNumber !== undefined) error.lineNumber = location.lineNumber;
-    if (location.columnNumber !== undefined) error.columnNumber = location.columnNumber;
-    errors.push(error);
-  });
-
-  page.on("pageerror", (error) => {
-    errors.push({
-      kind: "pageerror",
-      message: error instanceof Error ? error.message : String(error),
-    });
-  });
-
-  return errors;
 }
 
 function buildTestResults(
--- a/evals-cli/src/evaluator/browserEvaluator.ts
+++ b/evals-cli/src/evaluator/browserEvaluator.ts
@@ -5,7 +5,7 @@
 
 import { Browser, Page } from "puppeteer-core";
 import { WebmcpConfig } from "../types/config.js";
-import { Eval, TestResult, TestResults } from "../types/evals.js";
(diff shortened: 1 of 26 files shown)


evals-cli/src/evaluator/browserEvaluator.ts
```diff
@@ -119,13 +124,38 @@ async function runSingleBrowserTest(
...
-  const page = await browser.newPage();
+async function setupBrowserPage(page: Page, url: string): Promise<void> {
   await page.goto(url, {
     waitUntil: "networkidle2",
     timeout: 30000,
   });
-  return page;
+}
+
+export function collectBrowserConsoleErrors(page: Page): BrowserConsoleError[] {
```
swissspidy: Collecting _all_ console errors in general, even ones unrelated to WebMCP tool calls, is potentially very noisy. If we could establish some kind of error attribution, that would be very helpful. This would allow answering the question whether an error occurred during an WebMCP tool call or not.

`BrowserToolRegistry.executeToolChecked` could be a good place for that: add event listener before tool call, remove it after.

Or if we want both: add event listener at the start. Before a tool call, store the current tool name & args somewhere, and attribute that to any logged errors. After tool call, clean up the current tool name again.

Would make for nicer reporting, too.

alectimison-maker: Addressed in b39b3dd. Collection now runs at the BrowserToolRegistry execution boundary: the shared listener is installed while tool calls are active and removed in finally. Each diagnostic includes the active tool name/args; overlapping calls produce one diagnostic listing all active calls as possible sources instead of duplicating or falsely attributing it. I also added coverage for warnings/background noise, overlap, HTML escaping, and cleanup on success/failure. npm test passes 153 tests with the existing Chrome-launch integration skip; lint and format checks pass.

swissspidy:
@alectimison-maker heya, would you mind refreshing the PR to resolve conflicts? We did some refactoring/renaming recently.

alectimison-maker:
Rebased onto the latest `main` and resolved the evals refactor/rename conflicts. The browser-console diagnostics now use the current `BrowserToolRegistry` flow and retain per-tool-call attribution.

Validated with `npm test` (161 passed, 1 browser integration test skipped because no Chrome channel is available locally), `npm run lint` (0 errors; existing warnings only), and `npx oxfmt --check`.

swissspidy (approved):

PR #364, opened 2026-08-06
fix(evals): prevent 30s navigation timeout on valid null tool outputs

Kulikowski:
## What this PR does
This PR fixes a bug in [`evals-cli/src/evaluator/browser.ts`](file:///usr/local/google/home/kaskulikowski/Documents/workspaces/google-workspaces/webmcp-tools/evals-cli/src/evaluator/browser.ts) where tools returning `null` or `undefined` caused the test runner to freeze for 30 seconds.

## Why we need this
When a browser tool finished successfully and returned `null`, the code wrongly assumed the web page was opening a new URL. It called `waitForNavigation()`, which waited for 30 seconds until timing out with a `TimeoutError`.

## What changed
- **[`evals-cli/src/evaluator/browser.ts`](file:///usr/local/google/home/kaskulikowski/Documents/workspaces/google-workspaces/webmcp-tools/evals-cli/src/evaluator/browser.ts)**:
  - Updated tool return logic so `null` and `undefined` outputs return `"Success"` right away.
  - Removed the unsafe `waitForNavigation()` call that caused the 30-second hang.
  - Cleaned up event listeners for form tools so they unhook properly after running.
- **[`evals-cli/src/test/browserToolRegistry.test.ts`](file:///usr/local/google/home/kaskulikowski/Documents/workspaces/google-workspaces/webmcp-tools/evals-cli/src/test/browserToolRegistry.test.ts#L130-L150)**:
  - Updated unit tests to check that `null` tool outputs finish immediately without waiting for page navigation.

### DIFF
--- a/evals-cli/src/evaluator/browser.ts
+++ b/evals-cli/src/evaluator/browser.ts
@@ -50,7 +50,7 @@ export class BrowserToolRegistry implements ToolRegistry {
     let executionResult: { result?: any; error?: string } = {};
 
     try {
-      const tools = this.page.webmcp.tools() || [];
+      const tools = this.page.webmcp?.tools() || [];
       const tool = tools.find((t) => t.name === name);
       if (!tool) {
         return { success: false, error: `no tool named "${name}" was found` };
@@ -64,21 +64,33 @@ export class BrowserToolRegistry implements ToolRegistry {
           (resolve) => {
             let timer: NodeJS.Timeout | null = null;
 
+            const cleanup = () => {
+              if (timer) {
+                clearTimeout(timer);
+                timer = null;
+              }
+              this.page.webmcp?.off?.("toolinvoked", onToolInvoked);
+            };
+
             const onToolInvoked = (call: WebMCPToolCall) => {
               if (!call.tool || call.tool.name === name) {
+                if (timer) clearTimeout(timer);
                 timer = setTimeout(() => {
+                  cleanup();
                   resolve({ success: true, data: "pending form submission" });
                 }, 1000);
               }
             };
 
-            this.page.webmcp.once("toolinvoked", onToolInvoked);
+            const subscribe = this.page.webmcp?.on
+              ? this.page.webmcp.on.bind(this.page.webmcp)
+              : this.page.webmcp?.once?.bind(this.page.webmcp);
+            subscribe?.("toolinvoked", onToolInvoked);
 
             tool
               .execute(args)
               .then((res: WebMCPToolCallResult) => {
-                if (timer) clearTimeout(timer);
-                this.page.webmcp.off("toolinvoked", onToolInvoked);
+                cleanup();
                 if (res.status === "Completed") {
                   resolve({ success: true, data: res.output ?? "Success" });
                 } else if (res.status === "Error") {
@@ -91,10 +103,20 @@ export class BrowserToolRegistry implements ToolRegistry {
                 }
               })
               .catch((err: unknown) => {
-                if (timer) clearTimeout(timer);
-                this.page.webmcp.off("toolinvoked", onToolInvoked);
+                cleanup();
                 const message = err instanceof Error ? err.message : String(err);
-                resolve({ success: false, error: message });
+                if (
+                  message.includes("Execution context was destroyed") ||
+                  message.includes("Target closed") ||
+                  message.includes("navigating")
+                ) {
+                  resolve({
+                    success: true,
+                    data: `Tool ${name} executed and triggered a page navigation.`,
+                  });
+                } else {
+                  resolve({ success: false, error: message });
+                }
               });
           },
         );
@@ -111,7 +133,7 @@ export class BrowserToolRegistry implements ToolRegistry {
       } else {
         const res = await tool.execute(args);
         if (res.status === "Completed") {
-          executionResult.result = res.output !== undefined ? res.output : "Success";
+          executionResult.result = res.output ?? "Success";
         } else if (res.status === "Error") {
           return {
             success: false,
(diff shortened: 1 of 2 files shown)


evals-cli/src/evaluator/browser.ts
```diff
@@ -94,10 +106,20 @@ export class BrowserToolRegistry implements ToolRegistry {
                 }
               })
               .catch((err: unknown) => {
-                if (timer) clearTimeout(timer);
-                this.page.webmcp.off("toolinvoked", onToolInvoked);
+                cleanup();
                 const message = err instanceof Error ? err.message : String(err);
-                resolve({ success: false, error: message });
+                if (
```
9kubczas4: nit: I'm wondering, maybe would worth to extract simple function which will check this predicate and returns boolean? The same if statement is already used in this file.

evals-cli/src/evaluator/browser.ts
```diff
@@ -114,7 +136,7 @@ export class BrowserToolRegistry implements ToolRegistry {
       } else {
         const res = await tool.execute(args);
         if (res.status === "Completed") {
-          executionResult.result = res.output !== undefined ? res.output : "Success";
+          executionResult.result = res.output ?? "Success";
```
9kubczas4, line 138: idea: maybe we could define types for executionResult instead of assign string?

evals-cli/src/evaluator/browser.ts
```diff
@@ -124,15 +146,6 @@ export class BrowserToolRegistry implements ToolRegistry {
           return { success: false, error: `Tool execution status: ${res.status}` };
         }
       }
-
-      // If executionResult.result is null, it is due to a navigation happening.
```
swissspidy, line 124: We need to double check with demo apps that trigger navigation, e.g. the French Bistro site with toolautosubmit enabled. I think removing this breaks that use case.

evals-cli/src/evaluator/browser.ts
```diff
@@ -37,7 +37,7 @@ export class BrowserToolRegistry implements ToolRegistry {
   constructor(private page: BrowserPage) {}
 
   syncTools(): Tool[] {
-    const rawTools = this.page.webmcp.tools();
```
swissspidy: No need for all this optional chaining here or elsewhere. `this.page.webmcp` always exists

evals-cli/src/evaluator/browser.ts
swissspidy: Ah, I see how you check for the navigation error message above. Interesting!

So I assume you tested it with the demo app already?

evals-cli/src/evaluator/browser.ts
Kulikowski: Overall good idea! I think there are more places where we could introduce types so it might be worth to create a GitHub issue for that

evals-cli/src/evaluator/browser.ts
Kulikowski: image
image
Works with and without toolautosubmit

Kulikowski: To see this in action, use branch without those changes - then you can change any demo to return null, like 'add_topping' in pizza maker, and then run just 1 eval against it:

image

Kulikowski: It will pass, but adds 30s :)

swissspidy: Gotcha, thanks!

evals-cli/src/evaluator/browser.ts
```diff
@@ -64,21 +64,33 @@ export class BrowserToolRegistry implements ToolRegistry {
...
                   resolve({ success: true, data: "pending form submission" });
                 }, 1000);
               }
             };
 
-            this.page.webmcp.once("toolinvoked", onToolInvoked);
+            const subscribe = this.page.webmcp?.on
+              ? this.page.webmcp.on.bind(this.page.webmcp)
+              : this.page.webmcp?.once?.bind(this.page.webmcp);
+            subscribe?.("toolinvoked", onToolInvoked);
```
swissspidy:
```suggestion
            this.page.webmcp.on("toolinvoked", onToolInvoked);
```

evals-cli/src/evaluator/browser.ts
```diff
@@ -64,21 +64,33 @@ export class BrowserToolRegistry implements ToolRegistry {
           (resolve) => {
             let timer: NodeJS.Timeout | null = null;
 
+            const cleanup = () => {
+              if (timer) {
+                clearTimeout(timer);
+                timer = null;
+              }
+              this.page.webmcp?.off?.("toolinvoked", onToolInvoked);
```
swissspidy:
```suggestion
              this.page.webmcp.off("toolinvoked", onToolInvoked);
```

evals-cli/src/evaluator/browser.ts
```diff
@@ -50,7 +50,7 @@ export class BrowserToolRegistry implements ToolRegistry {
     let executionResult: { result?: any; error?: string } = {};
 
     try {
-      const tools = this.page.webmcp.tools() || [];
+      const tools = this.page.webmcp?.tools() || [];
```
swissspidy:
```suggestion
      const tools = this.page.webmcp.tools();
```

Kulikowski:
We are still investigating discrepancy between API and Puppeteer handling of outputs. This will inform if those changes are required. Please do not merge yet.

evals-cli/src/evaluator/browser.ts
9kubczas4: I'll create a ticket next week! Plus I will try code a bit as well :)

beaufortfrancois:
https://github.com/GoogleChromeLabs/webmcp-tools/pull/365/ should solve this issue.

PR #365, opened 2026-08-07
refactor(evals): remove null result navigation fallback

beaufortfrancois:
TIL `tool.execute()` can return null for non-navigations, so Puppeteer cannot reliably infer navigation from null outputs. 

Moreover, Puppeteer (using CDP) actually retrieves "automatically" the response after navigation. That's why I'm now removing the `waitForNavigation` handling on null results and clean up mock navigation tests.

/me happy!

### DIFF
--- a/evals-cli/examples/french-bistro/evals-toolautosubmit.json
+++ b/evals-cli/examples/french-bistro/evals-toolautosubmit.json
@@ -28,9 +28,13 @@
             "$contains": "anniversary"
           }
         },
-        "result": {
-          "$contains": "welcoming you"
-        }
+        "result": [
+          {
+            "@context": "https://schema.org",
+            "@type": "Message",
+            "text": "Hello Bob Smith,\nWe look forward to welcoming you on:\n\n2026-08-08 at 18:30\nParty of 2 • Terrace"
+          }
+        ]
       }
     ]
   },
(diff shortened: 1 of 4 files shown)


swissspidy (approved):
Nice

Kulikowski:
I think we should add a unit test to cover it, so it doesn't become broken in the future by accident

beaufortfrancois:
> I think we should add a unit test to cover it, so it doesn't become broken in the future by accident

What shall we test? Testing that `return null;` returns `null` is tested in Chromium and testing that getting the result from navigation is tested in CDP/Puppeteer.

Kulikowski:
> > I think we should add a unit test to cover it, so it doesn't become broken in the future by accident
> 
> What shall we test? Testing that `return null;` returns `null` is tested in Chromium and testing that getting the result from navigation is tested in CDP/Puppeteer.

What browserToolRegistry.test.ts actually tests is BrowserToolRegistry's own output formatting logic

I would see something like

```
  it("should preserve null return values in executeTool and executeToolChecked", async () => {
    const page = new MockBrowserPage();
    page.webmcp = {
      tools: () => [
        {
          name: "null_tool",
          description: "Tool that returns explicit null output",
          inputSchema: { type: "object" },
          execute: async () => ({
            status: "Completed",
            output: null,
          }),
        },
      ],
    };

    const registry = createRegistry(page);

    // 1. Verify executeToolChecked preserves null inside { success: true, result: null }
    const checkedResult = await registry.executeToolChecked("null_tool", {});
    assert.deepStrictEqual(checkedResult, { success: true, result: null });

    // 2. Verify executeTool returns null directly (instead of coercing to "Success")
    const result = await registry.executeTool("null_tool", {});
    assert.strictEqual(result, null);
  });
```

Kulikowski (approved):
LGTM, up to you regarding test, I'll close #364

beaufortfrancois:
Your test was more than useful as you can see! Thank you @Kulikowski! #teamwork

Kulikowski:
Teamwork 💪 LGTM

PR #366, opened 2026-08-07
Add image2svg to AWESOME_WEBMCP.md

botmonster:
Adds a single entry for [image2svg](https://botmonster.com/image2svg/) to the end of the Demos section: a free image to SVG converter with WebMCP tools, so an AI agent can convert images right on the page. The parent site botmonster.com also registers site-wide WebMCP tools for post search and sharing (live in the Chrome 149 origin trial).

I'll sign the CLA when the bot prompts.


beaufortfrancois:
Out of curiosity, how does the agent handle the data URI / raw base64 conversion?

PR #379, opened 2026-08-17
Rename evals-cli to webmcp-evals

beaufortfrancois:
This PR renames `evals-cli` folder to `webmcp-evals` to avoid confusion among developers who may assume this package evaluates more than WebMCP.

@Kulikowski Please review


Kulikowski (approved):
LGTM, I'll update shortlinks we have for this repo after its merged

PR #384, opened 2026-08-18
refactor: update tool execution to support object arguments

beaufortfrancois:
Updates tool execution for in-page agent demos to pass parsed input arguments to document.modelContext.executeTool, while retaining backward compatibility for versions of Chrome that still expect a JSON string.

References:
- https://chromium-review.googlesource.com/c/chromium/src/+/8250826
- https://github.com/webmachinelearning/webmcp/pull/246


Kulikowski (approved):
LGTM

PR #385, opened 2026-08-18
feat(evals): include browser console logs in reports

sudoeren:
## Summary

Closes #62

Browser evaluations can trigger JavaScript code on the target page that logs errors or warnings to the developer tools console. These failures were previously invisible in evals reports, making it hard to diagnose why an otherwise passing test misbehaved.

This PR captures browser console messages (`error`, `warn`, `assert`) and uncaught page errors (`pageerror`) during each browser eval and surfaces them in the HTML and JSON reports.

## Changes

- **types/evals.ts**: added `BrowserLogEntry` type and `browserLogs?: BrowserLogEntry[]` on `TestResult`.
- **evaluator/browserEvaluator.ts**: attach `console` and `pageerror` listeners on the test page before navigation, collect relevant messages per test run.
- **report/report.ts**: render a collapsible "Browser Console Logs" section per run, with error/pageerror entries highlighted in red and warnings in amber, plus a message/error count badge.
- **test/report.test.ts**: unit tests for rendering the logs section and omitting it when no logs were captured.

## Notes

- Only `error`, `warn`, `assert` console types and `pageerror` events are captured to keep report noise low.
- JSON reporter picks up `browserLogs` automatically since it serializes `TestResults`.
- Console logs are collected per page (one page per test), so all steps of a run share the same logs.

### DIFF
--- a/webmcp-evals/src/evaluator/browserEvaluator.ts
+++ b/webmcp-evals/src/evaluator/browserEvaluator.ts
@@ -4,7 +4,7 @@
  */
 
 import { WebmcpConfig } from "../types/config.js";
-import { Eval, TestResult, TestResults } from "../types/evals.js";
+import { BrowserLogEntry, Eval, TestResult, TestResults } from "../types/evals.js";
 import { ToolCall } from "../types/tools.js";
 import { countExpectedCalls, evaluateExecutionTrajectory } from "../utils.js";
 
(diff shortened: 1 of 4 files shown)


swissspidy:
Thanks for your PR. There's already https://github.com/GoogleChromeLabs/webmcp-tools/pull/361 for exactly this issue. What makes your PR different?

sudoeren:
Thanks, I just saw #361. Same issue, but the scope is a bit different. That PR captures only `error`-level messages while a tool call is active, whereas mine collects `error`/`warn`/`assert` and `pageerror` for the whole test run, without tool attribution. If you'd rather consolidate on #361, I'm happy to close mine, or adjust it to cover the parts #361 doesn't.

swissspidy:
Collecting all console errors in general, even ones unrelated to WebMCP tool calls, is potentially very noisy. That's why scope was reduced there on purpose. If someone wants to catch any error during tests, I recommend using a proper e2e test setup.

Let's consolidate on #361.

PR #388, opened 2026-08-23
Add WebMCP Today to Libraries & Tools

robertn702:
Adds [WebMCP Today](https://webmcp.today/) to `Libraries & Tools`.

WebMCP Today is an open-source package registry for discovering site-specific WebMCP packages and installing them with per-site install commands. Source: [robertn702/webmcp-today](https://github.com/robertn702/webmcp-today).


beaufortfrancois:
Thank you!

PR #405, opened 2026-09-02
Add three WebMCP Challenge entries and webmcp-kit

AshrafAhmed9:
Adding three apps built for the OpenAI WebMCP Challenge, plus a small library they share.

- **Cadence** — an issue tracker where the registered tool set changes live with selection and filters, and agent edits land on the same undo stack as a human's.
- **Consequence** — a high-stakes application form where fields declare who's allowed to fill them; the agent-facing tool refuses on human-only fields instead of failing quietly.
- **Relay** — a dispatch console with a build-time check enforcing that every mutating tool has a matching UI action and vice versa.
- **webmcp-kit** — the small typed library (`defineTool`, `useScopedTools`, `withConfirmation`, etc.) all three consume as a real dependency.

Note: I noticed the list already has an entry named "WebMCP Kit" (nekuda-ai/webmcp-kit) under Libraries & Tools. Mine is a separate, unrelated project that happens to share a similar name — happy to rename the link text if that's confusing, just let me know.


AshrafAhmed9:
recheck

AshrafAhmed9:
@googlebot I signed it!

beaufortfrancois (approved):
Thank you!

PR #421, opened 2026-09-04
Add Respira for WordPress to Demos

webmyc:
Adds a live WordPress demo to the Demos list.

The site registers a curated tool set on `document.modelContext` so an agent can read and edit the page it is looking at. Every write is staged on a duplicate of the page and needs the signed-in person to approve it before it reaches the published version, so the agent proposes and the human commits.

It runs on whatever page builder the site was already built with, which is the part that took the work: the tools write into the builder's own stored format rather than replacing the page with generated HTML.

- Demo: https://chatgpt-respira.instawp.site
- Source: https://github.com/respira-press/webmcp-for-wordpress (GPL-2.0)

Tested in ChatGPT's in-app browser, where it has been running in production since 2026-08-26.


beaufortfrancois:
Nice!

beaufortfrancois:
@webmyc I don't see any WebMCP tools in https://chatgpt-respira.instawp.site/ anymore. Is that expected?

webmyc:
> @webmyc I don't see any WebMCP tools in https://chatgpt-respira.instawp.site/ anymore. Is that expected?

@beaufortfrancois thanks for catching this, and no, not expected. Tool discovery on that site had been limited to signed-in users, and switching it to public exposed a second bug: signed-out visitors were still listed nothing. Both are fixed, and a signed-out visit now registers 30 tools. Anyone can list them; running anything still needs a signed-in user, and writes are staged for approval on the page.

PR #423, opened 2026-09-06
feat: Add goal-oriented simulation feature to webmcp-evals

9kubczas4:
## Summary

This draft explores a new goal-oriented eval type called `simulations`.

The existing eval format verifies whether an agent followed an expected tool-call trajectory.
This works well for deterministic cases, but can produce false negatives when different sequences
of valid actions lead to the same outcome.

A simulation instead defines:

- a user scenario,
- optional setup calls establishing the initial state,
- conversation limits,
- success criteria expressed as prose.

During execution, one model acts as the user, another as the agent under test, and a separate judge
evaluates the resulting transcript. Tool results are treated as evidence, while agent messages are
treated only as claims that require corroboration.

Simulations are exposed through a separate browser-only `simulate` command and produce
verdict-oriented console, JSON, and HTML reports. They complement the existing deterministic
`smoke` and trajectory-based evals rather than replacing them.

## Current status

This is intentionally a draft implementation. The code, prompts, error handling, and reporting
still need polishing, and some implementation paths are currently duplicated to keep the
experiment isolated from existing evaluators.

The implementation should nevertheless be good enough to evaluate the core idea and answer
whether goal-oriented, simulated-user evals are a useful direction for this project.

## Feedback wanted

In particular, I would like feedback on:

- whether simulations should remain a separate eval type,
- whether transcript-based judging provides sufficient evidence,
- whether the proposed `simulations.json` format is practical,
- whether the value justifies the additional cost and non-determinism of using three model roles.


webmcp-evals/src/commands/index.ts
```diff
@@ -274,6 +286,159 @@ export async function runSmokeCommand(options: CommandOptions, command?: Command
...
+    process.exit(1);
+  });
+
+  try {
+    const config: SimulationConfig = {
+      url,
+      simulationsFile,
+      model: opts.model,
+      ...(opts.judgeModel ? { judgeModel: opts.judgeModel } : {}),
+      ...(opts.userModel ? { userModel: opts.userModel } : {}),
```
Kulikowski, line 390: I am wondering if we should userModel optional and provide potential directions in the simulations.json file. Here we have potentially 3 different models: 1 running tools, 1 simulating user, 1 judge; Simulating user is imho great addition, but for simple goals it might be not needed overhead.

For the "judge" part, Maybe we could consider deterministic DOM check as well and keep LLM judge as an additional option?

Good direction!

webmcp-evals/src/simulate/toolSequence.ts
```diff
@@ -0,0 +1,207 @@
...
+      return undefined;
+    }
+  }
+
+  if (result === null || typeof result !== "object") return undefined;
+  const response = result as Record<string, unknown>;
+  if (
+    response.success === false ||
+    response.isError === true ||
+    (response.error !== undefined && typeof response.error === "string")
```
Kulikowski, line 115: nit: I would consider checking for empty error - Any tool returning error: "" on success will make every simulation against it fail. It seems that same should be added to smokeEvaluator.ts. WDYT?

webmcp-evals/src/commands/index.ts
```diff
@@ -274,6 +286,159 @@ export async function runSmokeCommand(options: CommandOptions, command?: Command
...
+  const url = opts.url!;
+  const simulationsFile = opts.simulations!;
+
+  process.on("SIGINT", () => {
+    console.log("\nGracefully shutting down from SIGINT (Ctrl-C)");
+    process.exit(1);
+  });
+
+  try {
+    const config: SimulationConfig = {
```
Kulikowski, line 385: + it seems that we don't handle --backend here as we have 3 possible configs - I would consider explicitly failing on backend param here or scope it just to local/browser evals instead of globally

webmcp-evals/src/commands/index.ts
```diff
@@ -274,6 +286,159 @@ export async function runSmokeCommand(options: CommandOptions, command?: Command
...
+      url,
+      simulationsFile,
+      model: opts.model,
+      ...(opts.judgeModel ? { judgeModel: opts.judgeModel } : {}),
```
Kulikowski, line 389: + I think we need to properly document roles of each model used here if we choose to keep all of them

webmcp-evals/src/evaluator/simulationEvaluator.ts
```diff
@@ -0,0 +1,314 @@
...
+    return {
+      simulation,
+      runIndex,
+      outcome: verdict.passed ? "pass" : "fail",
+      verdict,
+      ...(setupCalls ? { setupCalls } : {}),
+      conversation,
+      ...consoleErrors(),
+    };
+  } catch (error) {
```
Kulikowski, line 252: WDYT about returning whatever we collected here, like: conversation, setupCalls etc - if judge throws errors we wouldn't have it now right? It can help with debugging

webmcp-evals/src/evaluator/simulationEvaluator.ts
```diff
@@ -0,0 +1,314 @@
...
+        runIndex,
+        outcome: "error",
+        ...(setupCalls ? { setupCalls } : {}),
+        conversation,
+        error: `the conversation broke off: ${messageOf(conversation.error)}`,
+        ...consoleErrors(),
+      };
+    }
+
+    const verdict = await judgeOf(
```
Kulikowski, line 234: I would introduce the timeout on judge as well similar to the other model usages in simulation

webmcp-evals/src/report/simulationReport.ts
```diff
@@ -0,0 +1,388 @@
...
+type SimulationGroup = {
+  name: string;
+  runs: SimulationResult[];
+  passCount: number;
+};
+
+function groupBySimulation(results: SimulationResult[]): SimulationGroup[] {
+  const groups = new Map<string, SimulationGroup>();
+  for (const result of results) {
+    const name = result.simulation.name;
```
Kulikowski, line 41: nit: risk of name collision and unwanted merge?

webmcp-evals/src/simulate/userSimulator.ts
```diff
@@ -0,0 +1,79 @@
...
+  const message = text.split(END_CONVERSATION_MARKER).join("").trim();
+  return { message, done };
+}
+
+export async function simulateUserTurn(
+  request: UserTurnRequest,
+  model: LanguageModel,
+): Promise<UserTurnResult> {
+  const { system, messages } = buildUserSimulatorMessages(request);
+  const generated = await generateText({ model, system, messages });
```
Kulikowski, line 77: I think we need to use AbortController here so requests are being cancelled not only the timeout for waiting, same applies to other places in this codebase to be fair - (local/browser model usage)

Kulikowski: Other places are out of scope of this PR

webmcp-evals/src/simulate/conversation.ts
```diff
@@ -0,0 +1,194 @@
...
+            abortSignal: controller.signal,
+          }),
+          Math.max(remainingMs(), 1),
+          "the agent's turn",
+        );
+      } catch (thrown) {
+        // We are walking away from work still in flight; tell it to stop
+        // rather than leaving a browser driving itself in the background.
+        controller.abort();
+        throw thrown;
```
Kulikowski, line 152: Can we collect some useful data here ? Partial-steps and tool calls that already happened?

webmcp-evals/examples/commerce/shopping/simulations.json
```diff
@@ -0,0 +1,31 @@
+[
+  {
+    "name": "Find and buy a leather jacket",
+    "userScenario": "You need a jacket for a concert this weekend. You like leather, you have not decided on a specific one, and you want to complete the purchase today. Describe what you want the way a shopper would — do not mention product IDs, and do not tell the assistant which steps to take.",
```
Kulikowski, line 4: The separate schema makes sense as simulations judge outcomes, not trajectories. 

userScenario in the example files seems to restate behavior already covered by the userSimulatorSystemPrompt (like do not mention product IDs) - maybe we can skip it in scenario?

Kulikowski:
Very good direction! Let's chat about it!

PR #428, opened 2026-09-07
feat: add AbortSignal support to EvalTool execute

beaufortfrancois:
Starting in [Chrome 153.0.8009.0](https://chromiumdash.appspot.com/commit/e18e43744babcbe0d525dc204d18e7b5c36d6ddb), tool execution callbacks registered via `document.modelContext.registerTool()` receive an options object with an `AbortSignal` as their second argument (`execute(inputObject, { signal })`). See [webmachinelearning/webmcp#48](https://github.com/webmachinelearning/webmcp/issues/48) and [webmachinelearning/webmcp#247](https://github.com/webmachinelearning/webmcp/pull/247).

This PR adds `AbortSignal` cancellation support to the Web Worker execution in the maze demo's `eval_code` tool:

- **Bumps `webmcp-types`** to `^0.1.6` for `ToolExecuteCallbackOptions` typing.
- **Pre-aborted check**: If `signal?.aborted` is already set, execution immediately returns without spinning up the worker.
- **In-flight cancellation**: Listens for the `abort` event on `signal` to immediately terminate the worker, clear the timeout, revoke the blob URL, and resolve with an error message.
- **Cleanup**: Removes the `abort` event listener when execution completes naturally.
- **Bridge guard**: Prevents messaging the worker if the execution settled or aborted while an async `window.gameTools.executeTool` call was pending.

### DIFF
--- a/demos/webmcp-maze/package-lock.json
+++ b/demos/webmcp-maze/package-lock.json
@@ -17,7 +17,7 @@
         "typescript": "~5.9.3",
         "vite": "^8.0.16",
         "vitest": "^4.1.0",
-        "webmcp-types": "^0.1.3"
+        "webmcp-types": "^0.1.6"
       }
     },
     "node_modules/@emnapi/core": {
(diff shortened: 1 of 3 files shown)


andreban (approved):
LGTM

PR #403, opened 2026-09-01
Add timer to demonstrate new AbortSignal introduce in Chrome 153

marcianoskate:
In the upcoming Chrome 153, a change to the API introduces the ability to pass an `AbortSignal` to the tool registration and tool execution. This enables long-running tasks to be properly notified when the user or the agent interrupts the task.

PR #434, opened 2026-09-09
fix: pass options to registered tool execution in webmcp-polyfill

beaufortfrancois:
This PR passes the `options` parameter to `registeredTool._execute(parsedArgs, options)` in webmcp-polyfill.js so that execution options (e.g. abort signal for now) are properly forwarded to imperative tools as discovered in https://github.com/GoogleChromeLabs/webmcp-tools/pull/403/


andreban (approved):
LGTM

PR #447, opened 2026-09-11
refactor: pass inputObject in executeTool() with backwards compatibility

beaufortfrancois:
Following [webmachinelearning/webmcp#246](https://github.com/webmachinelearning/webmcp/pull/246), `executeTool()` now accepts an optional JavaScript object directly instead of a JSON string.

This PR updates `executeTool()` calls across the polyfill and demos to pass `inputObject` directly, while maintaining backwards compatibility for existing Chrome Stable versions that still require JSON stringified input:

```javascript
let result;
try {
  result = await document.modelContext.executeTool(tool, inputObject);
} catch (e) {
  // TODO: Remove when executeTool doesn't accept JSON stringified inputArgs in Chrome Stable.
  if (e.message.startsWith("Failed to parse input")) {
    result = await document.modelContext.executeTool(tool, JSON.stringify(inputObject));
  } else {
    throw e;
  }
}
```

### DIFF
--- a/demos/french-bistro/agent-worker.js
+++ b/demos/french-bistro/agent-worker.js
@@ -135,7 +135,7 @@ async function handleUserSubmit(text, port, requestId) {
             const tool = tools.find((t) => t.name == name);
             if (!tool) throw new Error(`Tool ${name} not found`);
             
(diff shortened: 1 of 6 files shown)


andreban (approved):
LGTM

PR #453, opened 2026-09-11
Revert "build(deps-dev): bump @eslint/js in /demos/analytics-dashboard"

swissspidy:
Reverts GoogleChromeLabs/webmcp-tools#356 as it breaks deployments. Main ESLint packages need to be updated at the same time.


andreban (approved):
LGTM

PR #471, opened 2026-09-15
fix(french-bistro): sanitize modalDetails HTML using DOMPurify

beaufortfrancois:
http://b/528741973


andreban (approved):
LGTM

The code does include an external dependency now, which it didn't before :(.

beaufortfrancois:
French bistro agent actually already depended on genai. See https://github.com/GoogleChromeLabs/webmcp-tools/blob/9cc745a544bc86bef819d8bcbe2461b49058fa9f/demos/french-bistro/agent.js#L195

PR #480, opened 2026-09-20
docs: add SitemapKit WebMCP demo

0nl1n1n:
## Summary

- add SitemapKit to the live WebMCP demos
- document its `find_sitemaps` and `check_sitemap` tools
- include a prompt that exercises both tools

I maintain SitemapKit. The linked page documents the integration, and both tools return real anonymous results before signup.

## Verification

- the linked page returns HTTP 200
- `git diff --check` passes
- SitemapKit was not already listed


0nl1n1n:
Closing because this documentation-only contribution requires a Google CLA that I cannot complete as part of this automated run. I will not ask maintainers to bypass it.

